PR #1, opened 2021-05-31
Remove outdated Jupyter Notebook

iqmtestd:
It does not work any more and it can be misleading to users.


jkotilahti:
For me the notebook works, assuming the following things:

- `jupyter` package is installed
- the `import cairosvg` and its usages are commented out 

Does it work for you after ensuring these things, or is there some other problem? If this works, then I would suggest doing the following changes instead of removing the notebook:

- Add `jupyter` requirement to `setup.py` `extras_require` as "notebooks" requirement.
- Remove or comment out the `cairosvg` related parts from the notebook

iqmtestd:
Thanks for the clarification. I had the impression that this is abandoned code. I was wrong.  I retract this PR.

In Ubuntu 20 my `jupyter` installation was incomplete at first, and after removing the `cairosvg` references it works with only minimal errors.

PR #16, opened 2022-06-03
[unitaryhack] Adding a post_build method to Element to avoid calls to super().build() when constructing chips.

### LINKED ISSUE #12: `super().build()` always needed in chips
At the end of each chip's `build()` method we must always remember to call `super().build()` because `Chip.build()` creates some geometry that is needed in all chips. It would be better if that was not needed.

Proposed solution is to use a new `post_build()` method that is always automatically called in `Element.produce_impl()`:

- [x] Add empty `post_build()` method to Element
- [x] Call `self.post_build()` in `Element.produce_impl()` after `self.build()`
- [x] Rename `Chip.build()` to `Chip.post_build()`
- [x] Remove `super().build()` call from all chips. 

The `post_build()` method of elements could later be used also for other actions that are needed after `build()`.

upsideon:
Fixes #12 by following the instructions provided in the issue:

- A `post_build` method has been added to the `Element` class which is called in `Element.produce_impl` following a call to `build`.
- The `Chip.build` method has been renamed to `Chip.post_build`.
- All `super().build()` calls have been removed from chips as well as the associated documentation for implementing new chips.

All tests pass locally, so chips should continue to be constructed properly following these changes.

### DIFF
--- a/docs/user_guide/python_workflow.rst
+++ b/docs/user_guide/python_workflow.rst
@@ -196,9 +196,6 @@ element created in the previous section into a new chip::
             self.insert_cell(cross_cell, pya.DTrans(half_width + 2000, half_width + 2000))
             self.insert_cell(cross_cell, pya.DTrans(half_width + 2000, half_width - 2000))
 
-            # Call the Chip-class build-method to produce the chip frame and possible ground plane grid.
-            super().build()
-
 This code can be copied to a new Python-file ``new_chip1.py`` in the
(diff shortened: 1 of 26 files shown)


iqmtestd:
The code looks OK to me. Everything works, let's merge it.

jkotilahti (approved):
Thanks, this is perfect! I will rebase/merge this now.

PR #17, opened 2022-06-17
[unitaryhack] Update GUI workflow for the current version of KLayout

### LINKED ISSUE #11: Point-and-click workflow tutorial outdated
With new KLayout and KQCircuits versions some things have changed so that parts of ["point-and-click" workflow tutorial](https://iqm-finland.github.io/KQCircuits/user_guide/gui_workflow.html) don't work as described or are confusing. At least:

- Part 4.4-5.4: Not easy to select qubit because waveguide cells are selected instead.
	* Alternative way to select a qubit is to drag a selection box around it. Pressing "q" can then be used to open the "object properties" window.
- Part 6.5: The button is now a magnifying glass instead of "...".
	* Can mention both, so that both older and new KLayout versions can be used
- Part 7.1: Ground grid is not visible by default, should tell to unhide it in the layers list.

The end result should be such that everything works when following exactly the instructions in the tutorial.

Renaud2002:
Hi, this resolves #11 as part of the unitaryhack issues. I have added the update suggested in the issue and the ones I found while going through the program.

### DIFF
--- a/docs/user_guide/gui_workflow.rst
+++ b/docs/user_guide/gui_workflow.rst
@@ -15,7 +15,10 @@ only after you know the API.
       for KLayout Viewer and KLayout Editor. On Linux you may have to go
       to *File -> Setup -> Application -> Editing Mode* and set *Use
       editing mode by default* to true, and restart KLayout.
-   #. In the *Libraries* panel on the left, choose *Chip Library* and drag
+      On macOS click on KLayout on the top right conner click prefrences 
+      on application click Editing Mode then click on use editing mode by default.
+      Then restart the KLayout program.
+   #. In the *Libraries* panel on the bottom left, choose *Chip Library* and drag
       and drop the *Demo* chip from there to the layout.
    #. Click *Ok* in the *PCell parameters* window that opens.
 
@@ -34,7 +37,7 @@ only after you know the API.
    #. In the *Layout* panel locate a charge line of the qubit. It
       connects the top left launcher and the qubit.
    #. Click once on the charge line to select it.
-   #. Press ``Shift+F2`` to zoom in.
+   #. Press ``Shift+F2`` to zoom in. 
    #. From the *Tools* panel above the *Layout* panel select a tool
       *Partial*.
    #. Double click on the center of the charge line. A new node should
@@ -47,7 +50,7 @@ only after you know the API.
    #. Press ``F2`` to see the whole top cell.
    #. Locate the qubit in the top left part of the chip.
    #. From the *Tools* panel select *Move*.
-   #. Click on the qubit.
+   #. Drag a selection box around the qubit.
    #. Click on a new location for a qubit.
 
 #. **Disable a coupling port for the qubit.**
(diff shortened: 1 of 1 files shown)


docs/user_guide/gui_workflow.rst
```diff
@@ -56,19 +59,21 @@ only after you know the API.
    #. Click once on the qubit.
    #. Click ``Shift+F2``
    #. Double-click on the qubit.
-   #. In the new *Object Properties* window locate a tab *PCell
+   #. In the new *Instance Properties* window locate a tab *PCell
       parameters*.
-   #. Scroll down in the list of parameters to locate *Coupler lengths*.
-   #. Change the value in the text box to ``100,160,0``.
+   #. Scroll down in the list of parameters to locate *Include coupler between qubits*.
+   #. Click on the box.
```
jkotilahti, line 65: Looks like in this part you changed it so that it changes a parameter of the Demo chip (`Include coupler between qubits`) instead of `Coupler lengths` parameter of a single qubit. It shouldn't be possible to change the demo chip parameters at this point if it was turned to static as explained in earlier step of the tutorial. Probably the confusion was because you didn't turn Demo chip to static and because the selection of the qubit does not work easily by double clicking. I suggest changing this part of the tutorial to the following:

```
#. **Disable a coupling port for the qubit.**

   #. From the *Tools* panel select *Select*.
   #. Drag a selection box around the qubit.
   #. Click ``Shift+F2``
   #. Press ``q``.
   #. In the new *Instance Properties* window locate a tab *PCell
      parameters*.
   #. Scroll down in the list of parameters to locate *Coupler lengths*.
   #. Change the value in the text box to ``100,160,0``.
   #. Click ``Ok``.
```

Please check if this makes sense to you and change accordingly.

docs/user_guide/gui_workflow.rst
```diff
@@ -56,19 +59,21 @@ only after you know the API.
...
    #. Click ``Ok``.
 
 #. **Create a new qubit.**
 
    #. Press ``F2`` to see the whole top cell.
    #. From the *Tools* panel select *Instance*.
-   #. In the new *Object Editor Options* window locate *Library*.
-   #. Using the drop-down menu select *Element Library*.
-   #. Click on a button ``...`` a little bit to the left from the
+   #. Double-click on the qubit.
```
jkotilahti, line 72: This line should be removed and the previous line should be changed to say `toolbar` instead of `*Tools* panel`.

docs/user_guide/gui_workflow.rst
```diff
@@ -56,19 +59,21 @@ only after you know the API.
...
+   #. In the new *Instance Properties* window locate *Library*.
```
jkotilahti, line 73:
```suggestion
   #. In the new *Editor Options* window locate *Library*.
```

This will be the name of the window if you click "Instance" from toolbar (instead of double-clicking existing qubit).

docs/user_guide/gui_workflow.rst
```diff
@@ -56,19 +59,21 @@ only after you know the API.
...
+   #. Using the drop-down menu select *Qubit Library*.
+   #. Click on a button ``🔍 `` a little bit to the left from the *Library*
```
jkotilahti, line 75:
```suggestion
   #. Click on a button ``🔍`` a little bit to the left from the *Library* 
```

jkotilahti (changes requested):
@Renaud2002 Thanks, mostly the changes look good! But there are some changes that should be done we can accept this PR, please see my comments above.

docs/user_guide/gui_workflow.rst
Renaud2002: @jkotilahti Hi, I am not so much which line you are saying should be removed do you mean the ```#. Double-click on the qubit```

jkotilahti: Yes, removing the `#. Double-click on the qubit` line is what I mean.

jkotilahti:
Thanks @Renaud2002, I think everything looks good now! You can find the CLA to sign here https://meetiqm.com/developers/clas/individual/ (it can also be found by following links in https://github.com/iqm-finland/KQCircuits#contributing). Note that it might take until next week before we can merge this if you didn't sign the CLA yet.

Renaud2002:
Hi @jkotilahti I just sign the QM CLA please let me know if you see it, thank you very much for all your help.

Renaud2002:
@jkotilahti Thank you very much for all your help through the process I am grateful.

jkotilahti (approved):
@Renaud2002 I received confirmation about the CLA and will merge this now. Thank you for contributing to KQCircuits!

PR #36, opened 2023-01-25
switch to pip3 if pip is not available in path

manzanillo:
Use subprocess to handle any error that comes from having pip not defined in path. In those cases, it will use pip3.

### DIFF
--- a/setup_within_klayout.py
+++ b/setup_within_klayout.py
@@ -37,7 +37,7 @@ Usage:
 
 import os
 import subprocess
-from sys import platform
+import sys
 from setup_helper import setup_symlinks, klayout_configdir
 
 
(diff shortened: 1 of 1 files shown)


qpavsmi (commented):
After testing and refining this fix I came up that the following solution works best:

```
   try:
       subprocess.check_output(f"pip install -r {pip_args}", shell=True)
   except subprocess.CalledProcessError as e:
       # Maybe pip utility goes by the name of pip3
       try:
           subprocess.check_output(f"pip3 install -r {pip_args}", shell=True)
       except subprocess.CalledProcessError as e2:
           print("Couldn't install required packages: pip and pip3 commands not found")
           sys.exit(1)
```
Also includes a user friendly error if both pip and pip3 doesn't work.
Please also change `from sys import platform` to `import sys` and there are two occurrences of `platform` that needs to be changed to `sys.platform`

qpavsmi:
@cla-bot check

PR #39, opened 2023-02-03
Use `importlib.util` for better support

### LINKED ISSUE #38: Simulations: module 'importlib' has not attribute 'util'
A user reported the following for the simple workload manager:
```
AttributeError: module 'importlib' has not attribute 'util'
```
This doesn't happen for me, but Stack Overflow would suggest using `import importlib.util` instead, so let's do that.

nikosavola:
Fix user reported bug for certain environments by being more specific in imports

Closes #38


caspar-iqm:
I think this change is technically required indeed. Even though I never got an error on this, the Pycharm code checker flags this issue on python 3.10.5.

I could imagine the following dynamic behavior: if `importlib.util` has ever been imported somewhere, it probably exists in the scope of `importlib` for the remainder of the python session.

qpavsmi (approved):
Changes make sense

PR #49, opened 2023-04-18
Add PyPI badge to the README

iqmtestd:


README.rst
nikosavola, line 1: Maybe we should replace the version tag badge with this as currently this looks a bit funny.

image

iqmtestd: Nothing funny, just repetition. I had the same though, but they are different things, first is the git tag second is the released PyPI package's tag. Ideally they are identical but they can be different too, for instance when PyPI release fails. Or when we push many things at once and the last one is not tagged (this is an other bug).

nikosavola: I think in the end we should assume the versions to be the same and have only the pypi badge, as it serves the purpose to provide a link to pypi as well. The releases page is anyway clickable from the main repo page w/o the badge.

nikosavola (approved):

PR #58, opened 2023-08-29
EdgePort's 'side' is inferred from 'towards'

### LINKED ISSUE #8: `produce_waveguide_to_port` does not infer side for EdgePort
I noticed that [`produce_waveguide_to_port`](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/simulation.py) does not infer the side to extend to when using `use_internal_ports=False`, i.e., EdgePort. It has to be given with the `side` param.
Supporting `side=None` to extend the simulation in the direction most indicated by `towards` is a natural choice I think.

iqmtestd:
in 'produce_waveguide_to_port()' if it is not already specified by the 'side' argument.

Fix #8


iqmtestd:
I've tested that all the generated simulation files remain identical to those generated from main.

qpavsmi (changes requested):
Good job on a straight-forward solution.

There are more instances of simulations producing a waveguide to port, not sure if you missed them or skipped them intentionally to include examples of explicit side definition. These simulations are:
- [ ] `hanger_resonator_sim.py` the two `produce_waveguide_to_port` calls below the ones you changed have explicit direction set
- [ ] `circular_capacitor_sim.py`
- [ ] `double_pads_sim.py`
- [ ] `finger_capacitor_sim.py`
- [ ] `klayout_package/python/scripts/simulations/smooth_capacitor_library_sim.py` (Note the directory)
- [ ] `klayout_package/python/scripts/simulations/internal_port_calibration.py`
- [ ] `klayout_package/python/scripts/simulations/stack_types_tutorial_sim.py`

My opinion is that these would be good to change for consistency, and we already have an example of explicit direction setting in `flip_chip_connector_sim.py`. But let me know if you have a reason why the code files in the list shouldn't be changed and I can approve this PR right away

iqmtestd:
> not sure if you missed them or skipped them intentionally

Intentionally :). Explicitly setting side is probably better. It kinda documents what's happening. I've only changed a few calls to illustrate the back-up mechanism implemented here.

> [In] `hanger_resonator_sim.py` the two `produce_waveguide_to_port` calls below the ones you changed have explicit direction set

Yes. These calls use the default `use_internal_ports=True`, hence `side` arguments are ignored. I was not trying to fix every unrelated issue (like removing unnecessary parameters) here.

qpavsmi (approved):
My request was resolved via agreement, will proceed to merging

PR #62, opened 2023-09-25
Reload existing PCells without closing KLayout

### LINKED ISSUE #14: Reload existing PCells without closing GUI
If I edit a PCell source code and want to see the changes within KLayout application, I need to:
1. run [Reload libraries](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/scripts/macros/reload.lym) macro
2. make a new copy of a PCell

Would be nice if there was one less manual step and I would not need to make a new PCell. Instead:
1. before reloading, the reload macro could extract the PCell information in the active layout 
2. removes the PCell instance
3. reloads the libraries
4. places instances of the PCells with same parameters and location

Maybe after this issue we should have the follow up features:

5. watchdog for code changes to autorefresh such that I would not even need to run refresh macro manually
6. partial refresh to refresh a single python module and corresponding PCell, not the full library (the changed PCell might be inside other PCells)

iqmtestd:
The "KQCircuits -> Reload libraries" command apart from reloading the libraries also saves and re-creates all PCells in every view using the updated code but retaining all the parameter changes of the user.

Fix #14

### DIFF
--- a/klayout_package/python/kqcircuits/util/layout_to_code.py
+++ b/klayout_package/python/kqcircuits/util/layout_to_code.py
@@ -20,7 +20,7 @@ import textwrap
 
 from kqcircuits.defaults import default_layers
 from kqcircuits.elements.chip_frame import ChipFrame
-from kqcircuits.elements.element import Element, get_refpoints
+from kqcircuits.elements.element import Element, get_refpoints, insert_cell_into
 from kqcircuits.elements.waveguide_coplanar import WaveguideCoplanar
 from kqcircuits.elements.waveguide_composite import WaveguideComposite, Node
 from kqcircuits.util.parameters import pdt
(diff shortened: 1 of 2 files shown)


iqmtestd:
I've converted back to draft because (apart from the trivial pylint error) there is still something wrong with this PR using Ubuntu 22 with the latest KLayout 0.28.12. It does work fine with Debian testing and its included KLayout 0.28.5 :)

iqmtestd:
> I've converted back to draft because (apart from the trivial pylint error) there is still something wrong with this PR using Ubuntu 22 with the latest KLayout 0.28.12. It does work fine with Debian testing and its included KLayout 0.28.5 :)

My mistake, there was no problem at all (only the fixed pylint issue), I just had some local configuration mishap.

iqmtestd:
Just rebased

qpavsmi (approved):
Works as advertised! This should make it a lot more convenient to do design work using KQCircuits

PR #66, opened 2023-10-04
Singularity build uses --fakeroot only when needed

iqmtestd:
Issue #1372


qpavsmi (approved):
Looks good, hope this fixes the CD job

Issue #91, opened 2024-05-02
Validate and report misconfigured simulation export


qpavsmi:
### Description

When running a simulation script, we need to define two sets of parameters: one that guides the shape of the geometry, and the other that guides the execution of the third-party simulator, such as Elmer or Ansys. Certain parameter settings of those two parameter sets can't work together. These should be validated and reported before the simulation export work is done.

Will provide concrete examples later. For the purposes of this issue we can implement a few misconfiguration scenarios but make the code architecture extendable so that adding new parameter set rules would be very easy.

### Use case

When I write a simulation export script in `klayout_package/python/scripts/simulations` I might notice or know better if I define simulation and solution parameters in such a way that they can't work together. Instead of spending too much time debugging the problem after the simulation work is done, I'd like the KQC simulation exporter to tell me that the parameters don't make sense and for what reason.

PR #94, opened 2024-05-30
Implement simulation object validation checks

rmoretti9:
This pull request addresses issue https://github.com/iqm-finland/KQCircuits/issues/91

### DIFF
--- a/klayout_package/python/kqcircuits/simulations/export/simulation_validate.py
+++ b/klayout_package/python/kqcircuits/simulations/export/simulation_validate.py
@@ -1,0 +1,119 @@
+# This code is part of KQCircuits
+# Copyright (C) 2024 IQM Finland Oy
+#
+# This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later
+# version.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied
+# warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with this program. If not, see
+# https://www.gnu.org/licenses/gpl-3.0.html.
+#
+# The software distribution should follow IQM trademark policy for open-source software
+# (meetiqm.com/iqm-open-source-trademark-policy). IQM welcomes contributions to the code.
+# Please see our contribution agreements for individuals (meetiqm.com/iqm-individual-contributor-license-agreement)
+# and organizations (meetiqm.com/iqm-organization-contributor-license-agreement).
+from kqcircuits.simulations.export.ansys.ansys_solution import (
+    AnsysEigenmodeSolution,
+    AnsysCurrentSolution,
+    AnsysVoltageSolution,
+    AnsysHfssSolution,
+)
+from kqcircuits.simulations.export.elmer.elmer_solution import ElmerVectorHelmholtzSolution, ElmerCapacitanceSolution
+
+
+class ValidateSim:
+    """Validation class that contains consistency checks."""
+
+    def has_no_ports_when_required(self, simulation, solution):
+        """Validation check: ensures that a simulation object has ports when the solution type requires it.
+        Args:
+            simulations: A Simulation object.
+            solutions: A Solution object.
+        Raises:
+            Errors when validation criteria are not met.
+        """
+        port_names = get_port_names(simulation)
+        sim_name = simulation.name
+        if not port_names and type(solution) in [
+            AnsysHfssSolution,
+            AnsysVoltageSolution,
+            AnsysCurrentSolution,
+            ElmerVectorHelmholtzSolution,
+            ElmerCapacitanceSolution,
+        ]:
+            raise ValidateSimError(
+                f"Simulation '{sim_name}' has no ports assigned. This is incompatible with {type(solution)}",
+                validation_type=self.has_no_ports_when_required.__name__,
+            )
+
+    def has_edgeport_when_forbidden(self, simulation, solution):
+        """Validation check: ensure that if at least one "EdgePort" is present, some solution types can't be chosen.
+        Args:
+            simulations: A Simulation object.
+            solutions: A Solution object.
+        Raises:
+            Errors when validation criteria are not met.
+        """
+        port_names = get_port_names(simulation)
+        sim_name = simulation.name
+        if "EdgePort" in port_names and type(solution) in [
+            AnsysEigenmodeSolution,
+            AnsysVoltageSolution,
+            AnsysCurrentSolution,
+        ]:
+            raise ValidateSimError(
+                f"Simulation '{sim_name}' has at least one 'EdgePort'. This is incompatible with {type(solution)}",
+                validation_type=self.has_edgeport_when_forbidden.__name__,
+            )
+
+    def flux_integration_layer_exists_if_needed(self, simulation, solution):
+        """Validation check related to the presence of layers and magnetic flux integration.
+        Args:
+            simulation: A Simulation object.
+        Raises:
+            Errors when validation criteria are not met.
+        """
+        sim_name = simulation.name
+        has_integrate_flux = hasattr(solution, "integrate_magnetic_flux")
+        integrate_flux = solution.integrate_magnetic_flux if has_integrate_flux else False
+
+        # Ensures that a layer with thickness == 0 and a material != "pec"
+        # exists in the setup when "integrate_magnetic_flux" is True.
+        if integrate_flux:
+            layers = simulation.layers
+            has_flux_integration_layer = False
+            for layer in layers.values():
+                if layer.get("thickness", -1) == 0 and layer.get("material", "") != "pec":
+                    has_flux_integration_layer = True
+                    break
+            if not has_flux_integration_layer:
+                raise ValidateSimError(
+                    f"Simulation '{sim_name}' has 'integrate_magnetic_flux = True' "
+                    + "but the integration layer is missing.",
+                    validation_type=self.flux_integration_layer_exists_if_needed.__name__,
+                )
+
+
+def get_port_names(simulation):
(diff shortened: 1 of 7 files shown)


rmoretti9:
I have now signed the CLA

qpavsmi:
Oh hi :D Thanks for submission! It will take me some time to receive confirmation that you CLA form went through but in the mean time I will go through the code and test that it works

qpavsmi:
@cla-bot check

qpavsmi (commented):
I would recommend to change the function signature of `validate_simulation` so that instead of taking a list of `Simulation`s, it takes a `Simulation` and a [`Solution`](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/export/solution.py) object, and handle them explicitly instead of addressing them by `simulation[0]` and `simulation[1]`. You can see for example in `elmer_export.py` if it detects that there is no `Solution` object paired with `Simulation`, then it will construct a `Solution` object from `solution_parameters`. But at that point we have already run `validate_simulation`. So maybe `validate_simulation` could instead be invoked at the beginning of the `for sim_sol in simulations` when we have explicitly defined `simulation` and `solution`.

`port_and_simulation_type_compatibility` could probably be separated into two cases: "If no ports, solution cant be X" and "if edgeport, solution cant be X". It's good that these validation functions have single responsibility, and its better to get a bigger list of clearly defined checks than a small list of checks that do subchecks. One `ValueError` return per function is probably a good rule of thumb. The check functions could also be separated into some distinct new file so they are more easily maintained.

You could maybe introduce a custom exception type instead of using `ValueError`, we might treat these errors somehow differently from generic errors in the future.

We would also like some unit tests for these checks. Since the checks don't rely on any real geometry but only on ports, you could probably mock a `Simulation` object without a `Cell` and just insert arbitrary ports in there.

rmoretti9:
Thank you @qpavsmi, I think I implemented your suggestions now, but I'm a bit confused about the unit tests since I have very little experience with them. I made a test module in `KQCircuits\tests\simulations\simulation_validate\test_simulation_validate.py` but I'm not sure whether this works properly or not. When I run `pytest --cov --cov-report term-missing -rf` this test fails for Empty Simulations because it tries to match it with `AnsysHfssSolution`

qpavsmi:
> Thank you @qpavsmi, I think I implemented your suggestions now, but I'm a bit confused about the unit tests since I have very little experience with them. I made a test module in `KQCircuits\tests\simulations\simulation_validate\test_simulation_validate.py` but I'm not sure whether this works properly or not. When I run `pytest --cov --cov-report term-missing -rf` this test fails for Empty Simulations because it tries to match it with `AnsysHfssSolution`

Took me some time to realise this, but it's actually not your test failing, but your pull request breaks one of the existing tests: [test_export_produces_output_files](https://github.com/iqm-finland/KQCircuits/blob/main/tests/simulations/ansys/ansys_export/test_ansys_export_write.py). It creates a completely empty simulation (no ports either), and checks that it can be exported without causing errors. Now if solution type is not specified, it defaults to `AnsysHfssSolution`. But now you validate that there must be a port if such solution type is used, hence the test fails.

It's best to modify this test so that it sets the solution type to something that passes this check. I found one way to fix this test is to fix the [fixtures](https://docs.pytest.org/en/6.2.x/fixture.html) we defined for simulation tests. Notice that in the failing unit test we give `perform_test_ansys_export_produces_output_files` as an argument and then call it like a function. Where does that come from? Well in `tests/conftest.py` we define the fixture which is a function that returns another function `perform_test_ansys_export_produces_output_implementation`.

As this is getting very complicated, I'll just say exactly what to do to fix this test. We would like to be able to set which solution type to use by calling `perform_test_ansys_export_produces_output_files(EmptySimulation, ansys_solution=AnsysSolution())` in [test_export_produces_output_files](https://github.com/iqm-finland/KQCircuits/blob/main/tests/simulations/ansys/ansys_export/test_ansys_export_write.py). To do that, we modify the fixture in `tests/conf.py`. The fixture itself takes other fixtures as arguments: `tmp_path` is a built-in fixture to get a path to a temporary directory, and `get_simulation` is defined above. But we should instead introduce the new `ansys_solution` argument into the inner function `perform_test_ansys_export_produces_output_implementation` that the fixture returns. So that piece of code would look something like:

```
@pytest.fixture
def perform_test_ansys_export_produces_output_files(tmp_path, get_simulation):
    def perform_test_ansys_export_produces_output_implementation(cls, ansys_solution=None, **parameters):
        simulation = get_simulation(cls, **parameters)
        bat_filename = export_ansys([(simulation, ansys_solution) if ansys_solution else simulation], path=tmp_path)
        assert Path(bat_filename).exists()

    return perform_test_ansys_export_produces_output_implementation
```
After this change the test should pass. Consider making similar change for existing elmer test fixtures that we have.

I would recommend to study the [fixtures](https://docs.pytest.org/en/6.2.x/fixture.html) since that could help you structure the tests you are writing better. Generally its good practice that each test function is written in such a way that they can be independently run, which means they set everything needed for the test (and tear it down if needed). Of course it gets annoying when the same set up is copy pasted for each unit test. This is where fixtures are useful, and pytest makes sure that the fixture is reconstructed from scratch for every unit test so that they are not made dependent on their order of execution. So in your example the initialisation of `MockSimulation` could be made into a fixture.

You will also need to test for negative cases: with this simulation and that solution I expect it to fail. You can do this using `with pytest.raises(ValidateSimError) as e:` block, google for examples or see [test_chip_coordinate_conversion.py](https://github.com/iqm-finland/KQCircuits/blob/main/tests/masks/mask_layout/test_chip_coordinate_conversion.py)

qpavsmi:
Remember to run `python -m black -l 120 -t py39 .` once you are close to being done

rmoretti9:
Hi @qpavsmi, thanks for the useful suggestion. I fixed the `perform_test_ansys_export_produces_output_files` fixture and implemented a few tests (two for every sim validation check). Let me know if I can improve it further

qpavsmi:
I think after addressing this round of suggestions we can approve and get this merged :)

rmoretti9:
> I think after addressing this round of suggestions we can approve and get this merged :)

Thanks @qpavsmi, I squashed my commits and marked this Pull Request as ready for review

qpavsmi:
> Thanks @qpavsmi, I squashed my commits and marked this Pull Request as ready for review

I'm not sure if its visible to you but there are many pending review points that I outlined based on the code changes 
image

These would be good to address before we merge

rmoretti9:
> > Thanks @qpavsmi, I squashed my commits and marked this Pull Request as ready for review
> 
> I'm not sure if its visible to you but there are many pending review points that I outlined based on the code changes image
> 
> These would be good to address before we merge

It seems like I cannot access these comments (or I don't know how to do that), but I modified the code based on the screenshot's suggestion. Except for the one about adding the empty new line. I'm not sure where it should be added.

qpavsmi (commented):
Should be visible now

qpavsmi:
> It seems like I cannot access these comments (or I don't know how to do that), but I modified the code based on the screenshot's suggestion. Except for the one about adding the empty new line. I'm not sure where it should be added.

You should be able to see the suggestions now I think: https://github.com/iqm-finland/KQCircuits/pull/94#pullrequestreview-2094009560

rmoretti9:
I applied all the suggested changes. Maybe the usage of @pytest.mark.parametrize in tests is a bit redundant as it is now, but it allows adding new tests quite easily

qpavsmi (dismissed):
Thanks for sticking through the review process :D Implementation is very clear now and the unit tests are comprehensive. Approving now, will merge after I have run some local tests.

klayout_package/python/kqcircuits/simulations/export/simulation_validate.py
```diff
@@ -0,0 +1,117 @@
...
+        Args:
+            simulation: A Simulation object.
+        Raises:
+            Errors when validation criteria are not met.
+        """
+        sim_name = simulation.name
+        has_integrate_flux = hasattr(solution, "integrate_magnetic_flux")
+        integrate_flux = solution.integrate_magnetic_flux if has_integrate_flux else False
+
+        # Ensures that a layer with thickness == 0 and a material != "pec" exists in the setup when "integrate_magnetic_flux" is True.
```
qpavsmi, line 83: CI linter found that this line is too long

klayout_package/python/kqcircuits/simulations/export/simulation_validate.py
```diff
@@ -0,0 +1,117 @@
...
+        if integrate_flux:
+            layers = simulation.layers
+            has_flux_integration_layer = False
+            for layer in layers.values():
+                if layer.get("thickness", -1) == 0 and layer.get("material", "") != "pec":
+                    has_flux_integration_layer = True
+                    break
+            if not has_flux_integration_layer:
+                raise ValidateSimError(
+                    f"Simulation '{sim_name}' has 'integrate_magnetic_flux = True' but the integration layer is missing.",
```
qpavsmi, line 93: also this line was found to be too long

qpavsmi (dismissed):
Approving again after fixing linter issues

klayout_package/python/kqcircuits/simulations/export/simulation_validate.py
```diff
@@ -0,0 +1,119 @@
...
+            layers = simulation.layers
+            has_flux_integration_layer = False
+            for layer in layers.values():
+                if layer.get("thickness", -1) == 0 and layer.get("material", "") != "pec":
+                    has_flux_integration_layer = True
+                    break
+            if not has_flux_integration_layer:
+                raise ValidateSimError(
+                    f"Simulation '{sim_name}' has 'integrate_magnetic_flux = True' 
+                    but the integration layer is missing.",
```
qpavsmi, line 95:
```suggestion
                    f"Simulation '{sim_name}' has 'integrate_magnetic_flux = True' "
                    + "but the integration layer is missing.",
```
This still fails documentation building for some reason, maybe change it to this?

qpavsmi: This actually fails many things :) But fixing it as I suggested should work

qpavsmi (approved):

PR #113, opened 2025-06-07
General rationale and first viable test

jbolns:
The commits on this PR aim to (and will) enable automated visualisation of simulations using ParaView in a way that is easily applicable across simulations.

The key driving idea:
- You can't pass arguments to a macro running in ParaView, but you can slightly modify the macro before feeding it to ParaView so that it always reflects the specific needs of a given sweep (and this turned out the be surprisingly achievable). 

Accordingly, the flow of actions goes as follows:
- A template macro exists
- When the simulation runs, the macro is set up with parameters applicable for all sweeps.
- Macro (and an additional file used to set a clean "state") is copied over to the simulation outputs.
- The paths to data_files are updated on each sweep.
- On each sweep, paraview is called using a clean state and a macro specific to the sweep.

The first commit contains:
- General approach for simulations producing pvtu files.
- Test using the `concentrictransmon_couplings_sim.py` simulation, which already works end to end. 

Subsequent commits will:
- Implement the approach for simulations producing vtu files.
- Test the approach with other simulation files.

### DIFF
--- a/klayout_package/python/scripts/simulations/concentrictransmon_couplings_sim.py
+++ b/klayout_package/python/scripts/simulations/concentrictransmon_couplings_sim.py
@@ -35,7 +35,10 @@ from kqcircuits.util.export_helper import (
     get_active_or_new_layout,
     open_with_klayout_or_default_application,
 )
+
 from kqcircuits.simulations.single_element_simulation import get_single_element_sim_class
+from kqcircuits.util.export_helper import setup_paraview_macro
+
 
 # Prepare output directory
 dir_path = create_or_empty_tmp_directory(Path(__file__).stem + "_output")
@@ -73,7 +76,7 @@ elmer_export_parameters = {
         "run_gmsh_gui": False,  # open gmsh gui after meshing
         "run_elmergrid": True,
         "run_elmer": True,
-        "run_paraview": False,  # opens results in ParaView after finishing
+        "run_paraview": True,  # opens results in ParaView after finishing
         "n_workers": 1,  # workers for first-level parallelisation
         "gmsh_n_threads": 4,  # -1 means all the physical cores
         "elmer_n_processes": 4,  # processes for second-level parallelisation
@@ -89,6 +92,9 @@ elmer_export_parameters = {
         "1t1_gap&1t1_signal_3": [4.0, 8.0],
         "1t1_gap&1t1_ground": [4.0, 8.0],
     },
+    # "paraview_automation": {
+    #    "enable": False,
+    # }
 }
 
 # Get layout
@@ -119,8 +125,20 @@ simulations += cross_sweep_simulation(
     },
 )
 
+# Automation parameters for ParaView visualisation
+paraview_settings = {
+    "enable": True,
+    "threshold": [1, 1],
+    "threshold_scalars": ["CELLS", "GeometryIds"],
+}
+
+if paraview_settings["enable"] == True:
+    setup_paraview_macro(**paraview_settings)
+
+
 # Export simulation files
 export_elmer(simulations, **elmer_export_parameters)
 
 # Write and open OAS file
-open_with_klayout_or_default_application(export_simulation_oas(simulations, dir_path))
+# open_with_klayout_or_default_application(export_simulation_oas(simulations, dir_path))
+export_simulation_oas(simulations, dir_path)
--- a/klayout_package/python/scripts/simulations/elmer/run_helpers.py
+++ b/klayout_package/python/scripts/simulations/elmer/run_helpers.py
@@ -331,14 +331,59 @@ def run_elmer_solver(json_data: dict[str, Any], exec_path_override: Path | str |
 
 def run_paraview(result_path: Path | str, n_processes: int, exec_path_override: Path | str | None = None):
     """Open simulation results in paraview"""
+
     paraview_executable = shutil.which("paraview")
     if paraview_executable is not None:
         if n_processes > 1:
             pvtu_files = glob.glob(f"{result_path}*.pvtu")
-            subprocess.check_call([paraview_executable] + pvtu_files, cwd=exec_path_override)
+            try:
+                # Try opening paraview with fully rendered visualisations
+                paraview_state_path, paraview_macro_path = adapt_paraview_macro(pvtu_files)
+                subprocess.check_call(
+                    [paraview_executable] + [paraview_state_path.__str__()] + [paraview_macro_path.__str__()]
+                )
+            except (ValueError, NameError, TypeError, SyntaxError) as e:
+                print(
+                    "ERROR WITH AUTOMATED PARAVIEW APPROACH. OPENING PARAVIEW WITH BASE DATA. YOU WILL NEED TO CREATE THE VISUALISATIONS MANUALLY."
+                )
+                # If automated approach errors out, open paraview with data files only
+                subprocess.check_call([paraview_executable] + pvtu_files, cwd=exec_path_override)
         else:
             vtu_files = glob.glob(f"{result_path}*.vtu")
             subprocess.check_call([paraview_executable] + vtu_files, cwd=exec_path_override)
     else:
         logging.warning("Paraview was not found! Make sure you have it installed: https://www.paraview.org/")
         sys.exit()
+
+
+def adapt_paraview_macro(data_files: list[str]):
+    """
+    Adapts paraview's foundational macro file to the specific data files in current simulation,
+    mainly by writting paths and name of data_files to the macro document.
+
+    """
+
+    # Key paths
+    root_folder_path = Path(Path.cwd())
+    paraview_state_path = Path(root_folder_path / "scripts/paraview_state.py")
+    paraview_macro_path = Path(root_folder_path / "scripts/paraview_macro.py")
+
+    # Adapt macro the specifics of this sweep
+    with open(paraview_macro_path, "r") as f:
+        lines = f.readlines()
+        f.close
+
+    new_lines = []
+    for line in lines:
+        if line.strip().startswith("data_folder ="):
+            new_lines.append("".join(["data_folder = ", '"', str(root_folder_path), '"', "\n"]))
+        elif line.strip().startswith("data_files ="):
+            new_lines.append("".join(["data_files = ", str(data_files), "\n"]))
+        else:
+            new_lines.append(line)
+
+    with open(paraview_macro_path, "w") as f:
+        f.writelines(new_lines)
+        f.close
+
+    return paraview_state_path, paraview_macro_path
--- a/klayout_package/python/scripts/simulations/post_process/paraview_macro.py
+++ b/klayout_package/python/scripts/simulations/post_process/paraview_macro.py
@@ -1,0 +1,79 @@
+# WHAT IS THIS FILE?
+# This file contains a .py "macro" that runs from within ParaView to automatically load and format of data files created during simulations.
+
+# WHY DOES THIS FILE TRIGGER "PYTHON ANALYSIS" WARNINGS WHEN OPENED IN VSCODE OR SIMILAR SOFTWARE?
+# This file uses dependencies that are NOT KQCircuits dependencies because it does NOT run within the KQCircuits environment.
+# The file is copied over to an output folder and ultimately runs from within a ParaView window.
+# ParaView has an in-built Python environment with its own packages.
+#
+# If you use VSCode, you can disable python analysis warnings by adding the following to your `.vscode/settings.json`:
+#   ```
+#   {
+#       ...
+#           "python.analysis.ignore": [ "**/paraview_macro.py", "**/paraview_state.py"]
+#       ...
+#   }
+#   ```
+
+
+# IMPORTS
+from paraview.simple import *
+
+paraview.simple._DisableFirstRenderCameraReset()
+
+# RE-WRITABLE VARIABLES
+# Since you cannot pass variables to a ParaView macro,
+# the variables below are written/re-written externally before running the Macro inside ParaView.
+data_folder = ""
+data_files = []
+kwargs = {"enable": True, "threshold": [1, 1], "threshold_scalars": ["CELLS", "GeometryIds"]}
+
+# LOAD DATA FILES
+registration_names = [f.rsplit("/")[1].split(".")[0] for f in data_files]
+for i in range(len(registration_names)):
+    reader = XMLPartitionedUnstructuredGridReader(
+        FileName=[data_folder + "/" + data_files[i]], registrationName=registration_names[i]
+    )
+    reader.TimeArray = "None"
+
+# INIT VIEW
+render_view = GetActiveViewOrCreate("RenderView")
+
+# FORMAT DATA
+for i in range(len(registration_names)):
+    parent = FindSource(registration_names[i])
+    parent.TimeArray = "None"
+    display = Show(parent, render_view, "UnstructuredGridRepresentation")
+    display.Representation = "Surface"
+    display.SetScalarBarVisibility(render_view, True)
+    Hide(parent, render_view)
+
+# ADD THRESHOLDS
+for i in range(len(registration_names)):
+    parent = FindSource(registration_names[i])
+    threshold = Threshold(registrationName=f"{registration_names[i]}-threshold", Input=parent)
+    threshold.ThresholdMethod = "Between"
+    threshold.LowerThreshold = kwargs["threshold"][0]
+    threshold.UpperThreshold = kwargs["threshold"][1]
+    threshold.Scalars = kwargs["threshold_scalars"]
+    threshold_display = Show(threshold, render_view, "UnstructuredGridRepresentation")
+    threshold_display.Representation = "Surface"
+    threshold_display.SetScalarBarVisibility(render_view, True)
+    if i + 1 < len(registration_names):
+        Hide(threshold, render_view)
+
+# UPDATE COLOURS
+potentialLUT = GetColorTransferFunction("potential")
+potentialPWF = GetOpacityTransferFunction("potential")
+potentialTF2D = GetTransferFunction2D("potential")
+
+# UPDATE LAYOUT & VIEW
+layout = GetLayout()
+layout.SetSize(720, 650)
+render_view.ResetCamera(False, 0.9)
+render_view.Update()
+render_view.Set(
+    CameraPosition=[0.001, 0.001, 0.006455783139039206],
+    CameraFocalPoint=[0.001, 0.001, 0.00022500000000000005],
+    CameraParallelScale=0.0016126453422870138,
+)
--- a/.gitignore
+++ b/.gitignore
@@ -134,6 +134,9 @@ dmypy.json
 # Additions
 # --------------------------------------------------------------------------------
 
+# VSCode local settings
+.vscode
+
 # JetBrains IDEs
 .idea/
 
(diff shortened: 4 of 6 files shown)


jbolns:
A lingering question I have, @qpavsmi,  is whether the thresholds can be pre-known as to feed them like I do in the first commit. 

This wouldn't be a breaking change. Given the need, I can use RegEx to extract them from the output files. Everyone loves a good RegEx.

It's just simpler like it is presented on the first commit.

qpavsmi:
> * You can't pass arguments to a macro running in ParaView, but you can slightly modify the macro before feeding it to ParaView so that it always reflects the specific needs of a given sweep (and this turned out the be surprisingly achievable).

Remember that a paraview macro is capable of reading files in the file system, so if it knows the locations of key files, like some .jsons, it can "be fed" some parameters that it can work with to set thresholds and zoom out the camera an appropriate amount.

jbolns:
@qpavsmi 

This starts to look like a fun healthy competition. Me like.  
- Parsing strings. I'm good with ReGex. So, in theory, no problem. The question there is whether the respective strings do follow a similar pattern or if there are limited patterns? Else, it's bound to be a "hit and miss". 
- JSON. Are there such JSON's already being created? That's be so easy. But if they are not already being created, I'd still have to parse the strings, so it would be unncessary to create them.

qpavsmi:
> Parsing strings. I'm good with ReGex. So, in theory, no problem. The question there is whether the respective strings do follow a similar pattern or if there are limited patterns? Else, it's bound to be a "hit and miss".

It's the `.sif` file that gets produced in simulation output folder for each sweep, where you can parse which thresholds to use. You can assume that the formatting of `.sif` is robust and doesn't change. I wonder if regex will be enough, the threshold data is stored in multiline `Body ... End` blocks.

jbolns:
Second commit enables ParaView automation without a need for any change in the original simulation script other than enabling paraview. There is also a fix for the (or perhaps *a*) bug that caused some animations not to render any data when ParaView opened.

In essence, all data needed for automation is taken from:
- arguments sent to `run_paraview`,
- information in the `.sif`file,
- information in the `.json` file. 

ReGex defines thresholds:
- It is a triple pattern match.
- It uses "Body ([digit || digitdigit)" and "$ substrate_1(any number of characters up to end of line)" to kind of *sandwich* contents in between. If contents in between match central pattern "Target Bodies(digit)", then it takes all matches as valid (which allows calculating lower and upper limits).

Camera adjustments could use a little additional thinking. 
- They mostly work,
- Defaults to standard view in the event of errors,
- But still somewhat spotty.

Successful PVTU tests:
- `concentrictransmon_couplings_sim.py`
- `waveguides_sim_compare.py`
- `simple_workload_manager_example.py`
- `waveguides_sim_cross_section.py`
- `kqc sim circular_capacitor_epr_sim.py`

Worth noting, additionally. 
- The settings that define the final look can be changed easily
- I couldn't really decide what is the best rendering default mode.
  - If data and tresholds are both set to surface, some simulations will not show the thresholds.
  - If data is hidden and thresholds are shown as surface, some simulations look very empty.
  - The chosen setting is a little ugly across all simulations, but does work for all tests I ran.

klayout_package/python/scripts/simulations/post_process/paraview_macro.py
```diff
@@ -0,0 +1,152 @@
+# WHAT IS THIS FILE?
```
qpavsmi, line 1: Small note here that `post_process` folder is reserved for a special category of scripts. I'd suggest you move `paraview_macro.py` under `klayout_package/python/scripts/simulations/elmer` folder. Paraview is kind of coupled with the Elmer software because Ansys has its own visualising built in.

klayout_package/python/scripts/simulations/concentrictransmon_couplings_sim.py
```diff
@@ -73,7 +74,7 @@
         "run_gmsh_gui": False,  # open gmsh gui after meshing
         "run_elmergrid": True,
         "run_elmer": True,
-        "run_paraview": False,  # opens results in ParaView after finishing
+        "run_paraview": True,  # opens results in ParaView after finishing
```
qpavsmi, line 77: Revert changes in this file please if we decide merging this pull request

jbolns: Done locally. Downloaded original from `main` and copied over. Pushes on next commit.

klayout_package/python/scripts/simulations/post_process/paraview_macro.py
jbolns: Done locally. Pushes on next commit.

klayout_package/python/scripts/simulations/elmer/run_helpers.py
```diff
@@ -329,16 +330,87 @@ def run_elmer_solver(json_data: dict[str, Any], exec_path_override: Path | str |
...
+        if line.strip().startswith("data_folder ="):
+            new_lines.append("".join(["data_folder = ", '"', str(root_folder_path), '"', "\n"]))
+        elif line.strip().startswith("data_files ="):
+            new_lines.append("".join(["data_files = ", str(data_files), "\n"]))
+        elif line.strip().startswith("cross_section =") and cross_section:
+            new_lines.append("".join(["cross_section = ", str(True), "\n"]))
+        elif line.strip().startswith("thresholds =") and sif_files:
+            new_lines.append("".join(["thresholds = ", str([threshold_low, threshold_up]), "\n"]))
+        else:
+            new_lines.append(line)
```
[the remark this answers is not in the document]
jbolns: Done. Pushed.

With a massive caveat that I'm honestly curious about how the competing solution solves it.

On my computer, this renders as a pop-up window with the visualisation and *no* user controls, rather than a full ParaView window with all user controls. From everything I have read about it, one would need to use a pvserver to link a window with user controls and a pvpython process, but that's more the kind of thing used in parallel dynamic visualisations (unless that's the long term goal?).

If the other solution found a way to link the full paraview window (spawned with the paraview executable) with pvpython (spawned with pvpython executable) without a server, standing ovation.

I can try add pvserver if needed. Should be fun.

klayout_package/python/scripts/simulations/elmer/paraview_macro.py
```diff
@@ -0,0 +1,178 @@
...
+def load_data(data_files: list[str]):
+    """Loads the foundational data from .pvtu or .vtu files, and returns parameters useful for other functions."""
+
+    simulation_folder_path = str(Path.cwd())
+    registration_names = []
+    simulation_name = vtu_pvtu = new_box = None
+
+    if data_files:
+        simulation_name = "" if not data_files else data_files[0].split("/")[0]
+        registration_names = [f.rsplit("/")[1].split(".")[0] for f in data_files]
```
qpavsmi, line 24: I have my setup installed on native Windows (not WSL), so relying on `/` separator causes issues. Windows tends to return path strings using `\\` instead of `/`, so tweaking the logic in such way fixed it. I think encompassing the string in `Path` and using it's built-in functions you will be able to separate simulation_name such that it works cross-platform.

klayout_package/python/scripts/simulations/elmer/run_helpers.py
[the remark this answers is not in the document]
qpavsmi: Thanks for looking into this more thoroughly than I did in my initial research.

I was able to get the user controls in the competing PR by changing one line of code. I would recommend revisiting this comment of mine (https://github.com/iqm-finland/KQCircuits/pull/113#discussion_r2135652261) and use paraview instead of pvpython, and you would be able to reach the perfect solution.

In my opinion, both of your PRs are so mature now, that it would be slower now to do direct plagiarism by copy-pasting parts of code, but instead it is faster now to consult each others' approaches and make quick fixes in your branch. I believe this is in the spirit of open source development and is totally acceptable.

klayout_package/python/scripts/simulations/elmer/paraview_macro.py
jbolns: Seems doable. Off to Tuesday sauna (considering you're probably in Finland, no need to explain why this takes priority). Will adjust later and push in next commit.

jbolns:
All as per above in latest commit, plus a new ReGex.

A comment you made on the other PR suggests `Target Bodies` might change position within the block in the future. RegEx now acounts for this. Checked using both [ReGex 101](https://regex101.com/) and running simulations with different threshold requirements, i.e., circular_capacitor_epr_sim.py.

klayout_package/python/scripts/simulations/elmer/run_helpers.py
```diff
@@ -329,16 +330,33 @@ def run_elmer_solver(json_data: dict[str, Any], exec_path_override: Path | str |
     )
 
 
-def run_paraview(result_path: Path | str, n_processes: int, exec_path_override: Path | str | None = None):
+def run_paraview(
+    result_path: Path | str, n_processes: int, exec_path_override: Path | str | None = None, cross_section: bool = False
+):
     """Open simulation results in paraview"""
+
+    pvpython_executable = shutil.which("pvpython")
```
qpavsmi, line 338: I was wondering if there is a need now to check if `pvpython_executable` exists anymore since we only use paraview now

jbolns: No need whatsoever. Already deleted locally (both the check and the conditional below). Will run some tests locally just for good practice, Pushes on next commit.

klayout_package/python/scripts/simulations/elmer/run_helpers.py
```diff
@@ -329,16 +330,33 @@ def run_elmer_solver(json_data: dict[str, Any], exec_path_override: Path | str |
...
-            subprocess.check_call([paraview_executable] + vtu_files, cwd=exec_path_override)
+    if pvpython_executable or paraview_executable:
+        pvtu_files = glob.glob(f"{result_path}*.pvtu")
+        vtu_files = glob.glob(f"{result_path}*.vtu")
+        sif_files = glob.glob(f"{result_path}*.sif")
+        data_files = pvtu_files if pvtu_files else vtu_files
+        paraview_macro_path = Path(Path(Path.cwd()) / "scripts/paraview_macro.py")
+
+        try:
+            with tempfile.NamedTemporaryFile(mode="w", suffix=".py", encoding="utf-8") as f:
```
qpavsmi, line 348: The temp file system doesn't seem to work on my setup, which is native Windows. Perhaps the files should just be proper files output to simulation output folder. We already have all kinds of intermittent files generated there, so storing these miniscripts there won't hurt. Screenshot illustrates the kind of error I get. Notice that it does seem to somehow load the correct `pvtu` files.

image

jbolns: Oh, ok. Fair enough. Can change. 
Where to save file to (will delete anyway, but there's sometimes a preference on what folder to hit with the file even if briefly)?

qpavsmi: Just under the same folder where it creates the json files in simulation output, using same naming format, that should be fine. No need to delete anything afterwards. Again, we create so many files already without deleting. The scripts could be useful anyway to reuse to quickly load up Paraview for specific element sweep

image

klayout_package/python/scripts/simulations/elmer/paraview_macro.py
```diff
@@ -0,0 +1,153 @@
...
+
+def apply_thresholds(
+    registration_names: list[str], vtu_pvtu: str, cross_section: bool, thresholds: list[int], render_view: object
+):
+    """Formats thresholds with information from get_thresholds."""
+
+    if (vtu_pvtu == "pvtu" and thresholds) or not (cross_section and thresholds):
+        for i in range(len(registration_names)):
+            parent = FindSource(registration_names[i])
+            threshold = Threshold(registrationName=f"{registration_names[i]}-threshold", Input=parent)
```
qpavsmi, line 62:
```suggestion
        for i, registration_name in enumerate(registration_names):
            parent = FindSource(registration_name)
            threshold = Threshold(registrationName=f"{registration_name}-threshold", Input=parent)
```

klayout_package/python/scripts/simulations/elmer/paraview_macro.py
```diff
@@ -0,0 +1,153 @@
...
+
+def run_macro(data_files: list[str], sif_files: list[str], cross_section: bool):
+    """Calls other functions to run applicable parts of a ParaView visualisation."""
+
+    paraview.simple._DisableFirstRenderCameraReset()
+    render_view = GetActiveViewOrCreate("RenderView")
+
+    simulation_folder_path = str(Path.cwd())
+    registration_names = []
+    simulation_name = vtu_pvtu = new_box = None
```
qpavsmi, line 137:
```suggestion
    simulation_name, vtu_pvtu = None, None
```
new_box I think not used here?

klayout_package/python/scripts/simulations/elmer/run_helpers.py
```diff
@@ -329,16 +330,33 @@ def run_elmer_solver(json_data: dict[str, Any], exec_path_override: Path | str |
     )
 
 
-def run_paraview(result_path: Path | str, n_processes: int, exec_path_override: Path | str | None = None):
+def run_paraview(
+    result_path: Path | str, n_processes: int, exec_path_override: Path | str | None = None, cross_section: bool = False
```
qpavsmi, line 334:
```suggestion
    result_path: Path | str, exec_path_override: Path | str | None = None, cross_section: bool = False
```
I think n_processes is no longer used in this function

klayout_package/python/scripts/simulations/elmer/paraview_macro.py
```diff
@@ -0,0 +1,153 @@
+# This file contains a .py "macro" that runs from within ParaView.
+# Note. The file uses dependencies that are NOT KQCircuits dependencies because it runs from within ParaView.
+# If you use VSCode, you can disable python analysis warnings by adding the following to your `.vscode/settings.json`:
+#   ```
+#   { "python.analysis.ignore": [ "**/paraview_macro.py"] }
+#   ```
+
+import re
+import sys
```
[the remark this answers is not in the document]
jbolns: Gone. Pushes next commit.

klayout_package/python/scripts/simulations/elmer/paraview_macro.py
jbolns: Done. ...

klayout_package/python/scripts/simulations/elmer/paraview_macro.py
```diff
@@ -0,0 +1,153 @@
...
+        try:
+            sif_file_path = Path(Path(Path.cwd()) / sif_files[0])
+            with open(sif_file_path, "r") as f:
+                sif_contents = f.read()
+            pattern = r"(?<=Body\s)(\d*\n*)(.*\s){0,4}Target\sBodies\(\d*\).*(?=substrate_1.*\n)"
+            matches = re.findall(pattern, sif_contents)
+            thresholds = [int(matches[0][0]), int(matches[0][0]) + len(matches) - 1]
+        except (ValueError, NameError, TypeError, FileNotFoundError) as e:
+            thresholds = []
+    return thresholds
```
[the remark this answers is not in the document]
qpavsmi: This change and many others are not yet present in this branch. If you meant by `Done` that you have made the change locally and will push commit after getting other things fixed, then no pressure. But in case you thought you pushed this change already, just informing that the change is not available online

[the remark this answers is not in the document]
jbolns: Sorry for the lack of clarity, since I was using a standard format, I just encapsulated "Pushes on next commit" in the "..." 
Will push soon. Just one more change and a couple safety tests to go. 
Applies to any other where I used "..."
@qpavsmi

klayout_package/python/scripts/simulations/elmer/run_helpers.py
jbolns: Ok, this one's done. Just testing to ensure files are there and well-named. Pushing soon.

klayout_package/python/scripts/simulations/elmer/run.py
```diff
@@ -147,7 +147,7 @@
         run_elmer_solver(json_data, path)
 
     if workflow.get("run_paraview", False):
-        run_paraview(name.joinpath("capacitance"), elmer_n_processes, path)
+        run_paraview(path / name / name, elmer_n_processes, path, cross_section=True)
```
qpavsmi, line 150:
```suggestion
        run_paraview(path / name / name, path, cross_section=True)
```

klayout_package/python/scripts/simulations/elmer/run_helpers.py
```diff
@@ -329,16 +329,33 @@ def run_elmer_solver(json_data: dict[str, Any], exec_path_override: Path | str |
...
+        paraview_macro_path = Path(Path(Path.cwd()) / "scripts/paraview_macro.py")
+        try:
+            temp_filename = f"{result_path.name}_pv.py"
+            with open(temp_filename, "w", encoding="utf-8") as f:
+                f.write(
+                    f"import sys\n"
+                    f"sys.path.insert(0, '{paraview_macro_path.parent}')\n"
+                    f"from paraview_macro import run_macro\n"
+                    f"run_macro({data_files}, {sif_files}, {cross_section})\n"
+                )
```
qpavsmi:
```suggestion
        paraview_macro_path = repr(str(Path(Path(Path.cwd()) / "scripts/paraview_macro.py").parent))
        try:
            temp_filename = f"{result_path.name}_pv.py"
            with open(temp_filename, "w", encoding="utf-8") as f:
                f.write(
                    f"import sys\n"
                    f"sys.path.insert(0, '{paraview_macro_path}')\n"
                    f"from paraview_macro import run_macro\n"
                    f"run_macro({data_files}, {sif_files}, {cross_section})\n"
                )
```
I was still getting platform specific problems since it produces a python file whose contents have unescaped backslashes. I've debugged this now for you and this change should fix it and be more robust no matter the platform.

qpavsmi: Yes, since we're not removing these files afterwards

klayout_package/python/scripts/simulations/elmer/paraview_macro.py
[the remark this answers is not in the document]
jbolns: This one is up already.

klayout_package/python/scripts/simulations/elmer/run_helpers.py
jbolns: This one is also up.

jbolns:
Perfect. It's been fun, @qpavsmi ! I like your attention to detail.

Ps. I read somewhere in the forum that some maintainers are forgetting to assign the issue prior to merge, which is apparently required per the unitaryHack guidelines.

qpavsmi (approved):
`paraview_macro.py` is nicely implemented, working both for 3D and 2D simulations

jbolns:
Should I do anything about those failed checks, @qpavsmi ? Copy/paste the copyright headers elsewhere on new files? Include an exception for `paraview_macro.py` somewhere in a CI check?

qpavsmi:
> Should I do anything about those failed checks, @qpavsmi ? Copy/paste the copyright headers elsewhere on new files? Include an exception for `paraview_macro.py` somewhere in a CI check?

Maybe it's better if I do it. I will try to make it so all changes you made will be in a commit authored by you, and all fixing for checks will be made by me in separate commit

jbolns:
I agree that's best. Just asking in case. 
Super!

PR #114, opened 2025-06-08
SuperInductor junction for Fluxonium devices #unitaryHACK2025

ljcamargo:
This is an almost ready PR which implements the requested circuit. There is only missing some shadows, some debug while placing in the double pads and some code cleanup, but it's mostly working. As I'm not expert in the field I have some doubts like the ground metal shapes that I don't know if they should remain or need to be deleted and modified and other nano-scale details like in the wire/junctions crossings offset, etc. So, as the ticket suggested I'm sending this PR for early review although I might finish totally (I think) in a day or two. Also competing for #unitaryHACK2025.

Captura de pantalla 2025-06-08 021702
Captura de pantalla 2025-06-08 021714

### DIFF
--- a/klayout_package/python/kqcircuits/junctions/super_inductor.py
+++ b/klayout_package/python/kqcircuits/junctions/super_inductor.py
@@ -1,0 +1,420 @@
+# This code is part of KQCircuits
+# Copyright (C) 2022 IQM Finland Oy
+#
+# This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later
+# version.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied
+# warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with this program. If not, see
+# https://www.gnu.org/licenses/gpl-3.0.html.
+#
+# The software distribution should follow IQM trademark policy for open-source software
+# (meetiqm.com/iqm-open-source-trademark-policy). IQM welcomes contributions to the code.
+# Please see our contribution agreements for individuals (meetiqm.com/iqm-individual-contributor-license-agreement)
+# and organizations (meetiqm.com/iqm-organization-contributor-license-agreement).
+
+
+from math import sqrt
+from kqcircuits.pya_resolver import pya
+from kqcircuits.util.parameters import Param, pdt
+from kqcircuits.junctions.squid import Squid
+from kqcircuits.util.symmetric_polygons import polygon_with_vsym
+
+
+class Manhattan(Squid):
+    """The PCell declaration for a Manhattan style SQUID.
+
+    This SQUID has two distinct sub-types automatically selected by loop-area.
+    """
+
+    finger_overshoot = Param(pdt.TypeDouble, "Length of fingers after the junction.", 1.0, unit="μm")
+    include_base_metal_gap = Param(pdt.TypeBoolean, "Include base metal gap layer", True)
+    shadow_margin = Param(pdt.TypeDouble, "Shadow layer margin near the the pads", 1.0, unit="μm")
+    compact_geometry = Param(pdt.TypeBoolean, "Compact geometry for metal addition.", False)
+    separate_junctions = Param(pdt.TypeBoolean, "Junctions to separate layer", False)
+    offset_compensation = Param(pdt.TypeDouble, "Junction lead offset from junction width", 0, unit="μm")
+    mirror_offset = Param(pdt.TypeBoolean, "Move the junction lead offset to the other lead", False)
+    finger_overlap = Param(pdt.TypeDouble, "Length of fingers inside the pads", 0.2, unit="μm")
+    single_junction = Param(pdt.TypeBoolean, "Disable the second junction", False)
+
+    def build(self):
+        self.produce_manhattan_squid(top_pad_layer="SIS_junction")
+
+    def produce_manhattan_squid(self, top_pad_layer):
+        # geometry constants
+
+        big_loop_height = 10
+        loop_bottom_y = 1.5
+        self.metal_gap_top_y = 20 if self.compact_geometry else 26.5
+        self.width = 36 if self.compact_geometry else 38  # total width of junction layer
+        self.height = 17 if self.compact_geometry else 20.2  # total height of junction layer
+        bp_height = 5  # bottom pad height
+        tp_width = 10  # top pad width
+        small_loop_height = 5.2
+        h_pad_height = 1
+        v_pad_width = 2
+        pad_overlap = 1
+
+        # corner rounding parameters
+
+        rounding_params = {"rinner": 0.5, "router": 0.5, "n": 64}
+        # In order inner corner rounding radius, outer corner rounding radius, number of point per rounded corner
+
+        # convenience variables
+        delta_j = self.loop_area / (big_loop_height)  # junction distance, a.k.a. loop width
+        tp_height = self.height - loop_bottom_y - big_loop_height  # top pad height
+        bp_gap_x = -self.width / 2 + (self.width - delta_j) / 2  # bottom gap left edge x-coordinate
+        bp_gap_x_min = -self.width / 2 + 7  # fixed at minimum size
+        finger_margin = 1  # make hats brim this much wider for good finger connection
+
+        # adjust for small loop geometry
+
+        small_loop = tp_width > -bp_gap_x * 2
+        if small_loop:
+            bp_gap_x = bp_gap_x_min
+            delta_j = self.loop_area / small_loop_height
+            edge_height = 2.7
+
+        junction_shapes_top = []
+        junction_shapes_bottom = []
+        shadow_shapes = []
+
+        # create rounded bottom part and top parts
+
+        self.produce_contact_pads(
+            bp_height,
+            bp_gap_x,
+            tp_height,
+            big_loop_height,
+            junction_shapes_bottom,
+            rounding_params,
+            shadow_shapes,
+        )
+
+        # create rectangular junction-support structures and junctions
+
+        if small_loop:
+            self._make_junctions(
+                self.produce_tp_small(
+                    edge_height,
+                    small_loop_height,
+                    loop_bottom_y,
+                    delta_j,
+                    finger_margin,
+                    junction_shapes_top,
+                    top_pad_layer,
+                    junction_shapes_bottom,
+                    shadow_shapes,
+                    h_pad_height,
+                    v_pad_width,
+                    pad_overlap,
+                    tp_width,
+                    tp_height,
+                    rounding_params,
+                )[3],
+                loop_bottom_y,
+            )
+        else:
+            self._make_junctions(
+                self.produce_tp_large(
+                    delta_j,
+                    finger_margin,
+                    tp_height,
+                    junction_shapes_top,
+                    top_pad_layer,
+                    junction_shapes_bottom,
+                    shadow_shapes,
+                    tp_width,
+                    rounding_params,
+                    h_pad_height,
+                )[1],
+                bp_height,
+                finger_margin,
+            )
+
+        self._add_shapes(junction_shapes_bottom, "SIS_junction")
+        self._add_shapes(junction_shapes_top, top_pad_layer)
+        self._add_shapes(shadow_shapes, "SIS_shadow")
+        self._produce_ground_metal_shapes()
+        self._produce_ground_grid_avoidance()
+        self._add_refpoints()
+
+    # Create top junction pad  for smaller area
+
+    def produce_tp_small(
+        self,
+        edge_height,
+        small_loop_height,
+        loop_bottom_y,
+        delta_j,
+        finger_margin,
+        junction_shapes_top,
+        top_pad_layer,
+        junction_shapes_bottom,
+        shadow_shapes,
+        h_pad_height,
+        v_pad_width,
+        pad_overlap,
+        tp_width,
+        tp_height,
+        rounding_params,
+    ):
+        small_hat = [
+            pya.DPoint(-v_pad_width / 2, self.height - edge_height + pad_overlap),
+            pya.DPoint(-v_pad_width / 2, small_loop_height + loop_bottom_y + h_pad_height),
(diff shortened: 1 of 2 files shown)


ljcamargo:
just signed!

qpavsmi:
@cla-bot check

qpavsmi:
Looks good! It's nice that you can already switch the junction type for double pads qubit to use this new super inductor!

I'm getting following unexpected geometry when I try to draw the super inductor.

image

I open KLayout, then I drag-and-drop DoublePads from Qubit library, and change `junction_type` to SuperInductor. Also just drag-and-dropping SuperInductor causes same bug. If this is due to default parameter values causing such geometries, I would recommend assigning default parameter values such that they produce some valid shape. If it's something else, I will only be able to check the code more thoroughly on Monday.

ljcamargo:
Hi @qpavsmi I just made several new commits for this ticket. I think it's finished now, it nows fits nicely on the double pad, but I have some questions.

1. I wonder if the connection pads need to be preserved as they were in Manhattan or are they very specific for this circuit? Because this superinductor is quite larger and more complex than Manhattan so I'm not sure if it's the aim to fit everything between the pads (about 12um available) or do we have available the whole 22um area of the chip? 
2. Does the phase slip junction have shadows as the rest of the junctions? Also, as this is the point when both wires meet, it's likely that both shadows lay on the same side on this particular junction depending on the parameters, is this a concern?

Captura de pantalla 2025-06-08 170524
Captura de pantalla 2025-06-08 171035
Captura de pantalla 2025-06-08 171217

qpavsmi:
@ljcamargo 

Looks like this is almost done! Thanks for your contribution! I will start delving into the code shortly. Answering your questions above:

1. That's a good point and something I didn't take into account when writing this issue. I think for simplicity its best to get rid of the pads since I think their specific design was tuned for Manhattan junction. For now let's just plug the superinductor wires directly into the qubit design. We can always update afterwards if having pads makes more sense. So the inductor could have length of 22 µm to start with. Also hopefully the length is parametrizable (haven't checked yet :D)
2. Phase slip junction should have shadows too, please add them :D. Shadows laying on the same side I think is fine for this round of design considering the deadline. If it becomes a problem, we can fix it afterwards :D

If you're wondering why I don't make some of these requirements completely definitive, is because I don't have the entire expertise on all details of how this junction would be fabricated, being a software developer in this quantum hardware project :D In our process we would consult the fabrication team on finer details of the design before merging the designs into main branch. Since they are currently very busy working on other projects, I opt for making a first iteration of the design, then settling the fabrication details when fabrication team is available, most likely after the hackathon is over :) I think now you have almost everything needed for the first iteration :)

klayout_package/python/kqcircuits/junctions/super_inductor.py
```diff
@@ -0,0 +1,747 @@
...
+    finger_overlap = Param(pdt.TypeDouble, "Length of fingers inside the pads.", 1.0, unit="μm")
+    height = Param(pdt.TypeDouble, "Height of the junction element.", 22.0, unit="μm")
+    width = Param(pdt.TypeDouble, "Width of the junction element.", 22.0, unit="μm")
+    pad_height = Param(pdt.TypeDouble, "Height of the junction pad.", 6.0, unit="μm")
+    pad_width = Param(pdt.TypeDouble, "Width of the junction pad.", 12.0, unit="μm")
+    pad_to_pad_separation = Param(pdt.TypeDouble, "Pad separation.", 6.0, unit="μm")
+    pad_rounding_radius = Param(pdt.TypeDouble, "Rounding radius of the junction pad.", 0.5, unit="μm")
+    #x_offset = Param(pdt.TypeDouble, "Horizontal junction offset.", 0, unit="μm")
+    #offset_compensation = Param(pdt.TypeDouble, "Junction lead offset from junction width", 0, unit="μm")
+    #mirror_offset = Param(pdt.TypeBoolean, "Move the junction lead offset to the other lead", False)
```
qpavsmi, line 72: Commented out, can be removed

klayout_package/python/kqcircuits/junctions/super_inductor.py
```diff
@@ -0,0 +1,747 @@
...
+from kqcircuits.pya_resolver import pya
+from kqcircuits.util.parameters import Param, pdt
+from kqcircuits.junctions.junction import Junction
+from kqcircuits.util.symmetric_polygons import polygon_with_vsym
+
+
+class SuperInductor(Junction):
+    """The PCell declaration for a Manhattan style single junction."""
+
+    junction_length = Param(pdt.TypeDouble, "Junction length", 5, unit="μm")
```
qpavsmi, line 30: It seems that this parameter is not actually being used in the code, can't find `self.junction_length` keyword in the code. Changing this value in KLayout has no effect on superinductor shape, presumably it should change the horizontal length of the junction pieces, same way how `phase_slip_junction_length` works?

klayout_package/python/kqcircuits/junctions/super_inductor.py
```diff
@@ -0,0 +1,747 @@
...
+        pdt.TypeDouble, "Minimum height of the shadow layer.", 0.04, unit="μm"
+    )
+    finger_overshoot = Param(pdt.TypeDouble, "Length of fingers after the junction.", 1.0, unit="μm")
+    include_base_metal_gap = Param(pdt.TypeBoolean, "Include base metal gap layer.", True)
+    include_base_metal_addition = Param(pdt.TypeBoolean, "Include base metal addition layer.", True)
+    shadow_margin = Param(pdt.TypeDouble, "Shadow layer margin near the the pads.", 0.5, unit="μm")
+    separate_junctions = Param(pdt.TypeBoolean, "Junctions to separate layer.", False)
+    finger_overlap = Param(pdt.TypeDouble, "Length of fingers inside the pads.", 1.0, unit="μm")
+    height = Param(pdt.TypeDouble, "Height of the junction element.", 22.0, unit="μm")
+    width = Param(pdt.TypeDouble, "Width of the junction element.", 22.0, unit="μm")
```
qpavsmi, line 65: Changing these in KLayout don't seem to have effect, though these parameters do seem to be used in code. If they have no effect on inductor shape, could the parameters be removed and static numbers be used in their place, with a comment explaining where 22.0 value comes from?

klayout_package/python/kqcircuits/junctions/super_inductor.py
```diff
@@ -0,0 +1,747 @@
...
+class SuperInductor(Junction):
+    """The PCell declaration for a Manhattan style single junction."""
+
+    junction_length = Param(pdt.TypeDouble, "Junction length", 5, unit="μm")
+    wire_width = Param(pdt.TypeDouble, "Wire width", 0.02, unit="μm")
+    phase_slip_junction_length = Param(pdt.TypeDouble, "Phase slip junction length", 0.5, unit="μm")
+    squid_area_width = Param(pdt.TypeDouble, "Width of Squids Area.", 2, unit="μm")
+    squid_area_height = Param(pdt.TypeDouble, "Height of Squids Area.", 5, unit="μm")
+    squid_count = Param(pdt.TypeInt, "Number of Squids in the Super Inductor.", 8)
+    squid_junction_length = Param(pdt.TypeDouble, "Length of the SQUID junctions.", 0.25, unit="μm")
```
qpavsmi, line 36: `squid_junction_length` parameter name is a bit confusing because we reserve junction terminology for the "horizontal" bits, but here we are adjusting the length of "vertical" bits. I'd suggest `squid_wire_length` or `squid_distance`.

klayout_package/python/kqcircuits/junctions/super_inductor.py
```diff
@@ -0,0 +1,747 @@
...
+        def junction_shadow(width, height):
+            shadow_shape = pya.DBox(0, -self.shadow_min_height/2, self.shadow_width, self.shadow_min_height/2)
+            left_transform = pya.DTrans(0, False, -self.shadow_width, height/2)
+            right_transform = pya.DTrans(0, False, width, height/2)
+            return [
+                left_transform * shadow_shape,
+                right_transform * shadow_shape,
+            ]
+
+        # SQUID ARRAY - Corrected version
```
qpavsmi, line 146:
```suggestion
        # SQUID ARRAY
```

klayout_package/python/kqcircuits/junctions/super_inductor.py
```diff
@@ -0,0 +1,747 @@
...
+            )
+            [left, right] = junction_shadow(squid_width, self.junction_width)
+            shadow.insert(squid_lower_transform * left * squid_transform)
+            shadow.insert(squid_lower_transform * right * squid_transform)
+            shadow.insert(squid_upper_transform * left * squid_transform)
+            shadow.insert(squid_upper_transform * right * squid_transform)
+
+
+        # Squid Pylon Wires
+        for i in range(int(squid_count)):
```
qpavsmi, line 179:
```suggestion
        for i in range(squid_count):
```

klayout_package/python/kqcircuits/junctions/super_inductor.py
```diff
@@ -0,0 +1,747 @@
...
+        #self._add_shapes(junction_shapes_top, "SIS_junction")
+        #self._add_shapes(shadow_shapes, "SIS_shadow")
+        self._produce_ground_metal_shapes()
+        self._produce_ground_grid_avoidance()
+        self._add_refpoints()
+
+    def _make_super_inductor_junctions(self, offset):
+        layer_name = "SIS_junction_2" if self.separate_junctions else "SIS_junction"
+        shape_into = self.cell.shapes(self.get_layer(layer_name))
+        shadow = self.cell.shapes(self.get_layer("SIS_shadow"))
```
qpavsmi, line 129:
```suggestion
        junction_shapes = self.cell.shapes(self.get_layer(layer_name))
        shadow_shapes = self.cell.shapes(self.get_layer("SIS_shadow"))
```

klayout_package/python/kqcircuits/junctions/super_inductor.py
```diff
@@ -0,0 +1,747 @@
...
+        
+        last_tapper_north_step_count = tapper_steps # <- export
+        last_tapper_north_transform = tapper_north_transform * pya.DTrans(
+            0, False, 
+            adjusted_tapper_horizontal_displacement - tapper_unit_width, 
+            last_tapper_north_step_count * tapper_unit_height
+        ) # <- export
+        
+
+        # TAPPER South
```
qpavsmi, line 311: The `# TAPPER North` and `TAPPER South` code blocks seem to be slightly changed copy-paste code. Would you be able to encapsulate this code in a function, then call the same function twice with different arguments to get the north taper and south taper part? This would shorten the code and make it easier to follow. If you won't be able to make this in time, that's totally fine.

klayout_package/python/kqcircuits/junctions/super_inductor.py
```diff
@@ -0,0 +1,747 @@
...
+            shadow_transform = pya.DTrans(
+                0, False,
+                0, tower_unit_height - (self.junction_width)
+            )
+            shadow.insert(jsl  * tower_north_east_transform * tower_step_transform * shadow_transform)
+            shadow.insert(jsr * tower_north_east_transform * tower_step_transform * shadow_transform)
+
+
+        # ----
+        # Tower SouthWest (Inverted version of Tower NordWest)
```
qpavsmi, line 520: I wonder if `# Tower SouthWest` and `# Tower NordWest` parts can be implemented using same code function, perhaps with different arguments. We could reduce the entire code of `SuperInductor` by half if we could have a function that takes care of drawing one half of the inductor, then call the function a second time to draw the other half. Again, its fine to keep it as it is if you won't make it in time for deadline.

ljcamargo:
Hi @qpavsmi I made several commits covering all fixes suggested but I need your help checking the name of the parameters, because trying to follow the convention of the existing circuit I came to some names that are counter intuitive, so please in the image check it (specially squid_junction_length(vertical) vs junction_length(horizontal)). All the other measures are calculated and there are no other non-parameterized hardcoded valued. 

About the slit, I added the shadows, but as far as I can foresee, there will always be a place where the shadows does not alternate as the structure except for the squid array is vertically mirrored, but in case the alternation is very critical in the phase slip perhaps we can tolerate this non alternation in one of the bends but with the secondary effect that the circuit won't be vertically symmetrical.

Btw, I didn't want to make too much code abstraction until being very certain that it's correct, for example in this case, the halves may not end up being symmetrical, but very prone to do this abstraction afterwards.

squid_measure

qpavsmi:
Hi @ljcamargo 

We're approaching the deadline and in addition to reviewing other PRs, I've had a surge of work related tasks this week. But I can guarantee that the bounty will be assigned to you based on the work you've done already. I will tomorrow review your most recent changes and have a look at your parameter naming proposals. Except for the naming suggestions I outlined in last review round, I am totally fine with the namings. UH allows me to extend the deadline so don't worry if I proceed with merging this PR only after the deadline passes. Watch this space, there might be small tweaks I'll ask you to add to your code. Good job on this design!

ljcamargo:
Hello @qpavsmi the variables names have been changed. I waited until I specified them all in the drawing for you to check them out and rename them all together. Now I've renamed all vars as you suggested. And, I'm sorry, I noticed I had wrongly renamed all taper to "tapper" (I'm unfamiliar with this word  and happens that Tapper is a last name!), but now they are all renamed to taper. I hope it's now complete, I'm also updating the image with the new var names

klayout_package/python/kqcircuits/junctions/super_inductor.py
```diff
@@ -0,0 +1,734 @@
...
+class SuperInductor(Junction):
+    """The PCell declaration for a Manhattan style single junction."""
+
+    junction_length = Param(pdt.TypeDouble, "Junction length", 2, unit="μm")
+    wire_width = Param(pdt.TypeDouble, "Wire width", 0.02, unit="μm")
+    phase_slip_junction_length = Param(pdt.TypeDouble, "Phase slip junction length", 0.5, unit="μm")
+    squid_area_width = Param(pdt.TypeDouble, "Width of Squids Area.", 2, unit="μm")
+    squid_area_height = Param(pdt.TypeDouble, "Height of Squids Area.", 5, unit="μm")
+    squid_count = Param(pdt.TypeInt, "Number of Squids in the Super Inductor.", 8)
+    squid_wire_length = Param(pdt.TypeDouble, "Length of the SQUID junctions.", 0.25, unit="μm")
```
qpavsmi, line 35:
```suggestion
    squid_wire_length = Param(pdt.TypeDouble, "Length of the wire between SQUIDs.", 0.25, unit="μm")
```

qpavsmi (approved):
Very nice and parametrizable design of a fluxonium element

PR #115, opened 2025-06-11
Circular Transmon Single Island EPR Implementation

### LINKED ISSUE #109: Enable EPR calculation for Circular Transmon Single Island qubit
## Introductory note

> [!IMPORTANT]
> The description for this task is long and can look overwhelming on first glance. We assure you that the actual task to implement is not as demanding as you might think.

The task boils down to defining a small set of parametrized shapes using [KLayout API](https://www.klayout.de/doc-qt5/code/index.html) for existing element in KQCircuits. There are plenty of examples already implemented in KQCircuits you can use for reference.

The reason the issue description blew up is because the method used in KQCircuits is quite novel and there is no exact literature available for it (yet), so to some extent we explain it here from scratch. In ``First look`` chapter we outline steps on how to navigate KQCircuits GUI to access the parametrized shapes for other elements. We explain how the method is organized in KQCircuits code, and we provide a suggestion on how the shapes could be implemented to be eligible for the bounty. Finally, we guide the participant through the steps to run a simulation. Note that this last step of running the simulation is not at all required to be eligible for the bounty. We felt that removing any part of this description would just cause more confusion. Most of this description is background information that only needs to be read once.

So please trust us as we assure you that the implementation work of this issue is less demanding than reading the issue description. Please submit a PR early on, even when its not fully completed, as this is the best way we can guide you through the implementation process by providing early feedback.

## Prerequisites

Minimal requirement: KQCircuits [GUI installation](https://iqm-finland.github.io/KQCircuits/developer/setup.html)

Above is all that is needed to get the issue done. Running simulations is not a strict requirement to implement an acceptable submission, but it would motivate your accomplishment to see your implementation work in actual simulation. Below are additional requirements to set up a simulation environment:

- KQCircuits [standalone](https://iqm-finland.github.io/KQCircuits/developer/standalone.html)
  - Install core, `tests` and `simulations` requirements
  - We recommend using [venv](https://docs.python.org/3/library/venv.html) to contain requirements
- [Elmer FEM simulator](https://www.elmerfem.org/blog/binaries/). You can install binaries directly, but for linux (or windows WSL) we recommend using [Singularity](https://iqm-finland.github.io/KQCircuits/developer/containers.html#singularity-usage) so that you can pull an image with pre-installed Elmer, configured specifically for KQCircuits use.
- [Paraview](https://www.paraview.org/) to visualize simulation results

## Description

For this issue we take an existing qubit from KQC's qubit library, and we define additional geometry to enable metal-edge-corrected simulation method to calculate Energy Participation Ratio (EPR) for this qubit. We simulate the qubit as it is put into the first energy state to calculate how the energy is spread at different parts of qubit geometry. Exploring different shapes of the qubit given its parametrization, we can recalculate EPR for different qubit shapes. Shapes for which energy participation is concentrated somewhere other than the oxidized layers of the metal and substrate edges, tend to have higher T1, meaning that the qubit is more likely to keep its energy state longer, allowing to execute longer quantum algorithms in between measurements.

## First look

The qubit is called `Circular Transmon Single Island`. [This is the code for it](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/qubits/circular_transmon_single_island.py).

To showcase the qubit in KLayout, choose `Qubit Library` beneath the `Libraries` panel, then drag-and-drop `Circular Transmon Single Island` into the layout. Click to place the copies where you wish, then press ESC to stop placing the qubit. F2 to zoom out to qubit's size, `*` button to show full hierarchy of cells.

Image

> [!TIP]
> There is also a way to invoke KLayout from your favorite IDE such that with [qubit code](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/qubits/circular_transmon_single_island.py) open, it will open KLayout with single copy of element placed at the center of the layout. Read more about it [here](https://iqm-finland.github.io/KQCircuits/user_guide/file_system_and_utilities.html#opening-element-or-chip-from-an-ide).

Now try drag-and-dropping the `Circular Capacitor` element from `Element Library`, then double-click it. Choose `PCell parameters` and scroll down to see `_epr_*` parameters. Check the `_epr_show` parameter, then choose some layer for `_epr_cross_section_cut_layer` parameter (some layers are less visible than others, use `1b1_metal_gap` to be sure it becomes visible). Do the same for other `_epr_part_reg_*_layer` parameters, one at a time to not clutter the layout too much.

Image

You should see annotated lines at different parts of the capacitor, which are called "correction cuts". Then for each `_epr_part_reg_*_layer` parameter there is a shape covering a key region in the element, which are called "partition region". Notice that as you change other geometry parameters of the capacitor, the shapes will adjust themselves to fit the new shape. Your task is to implement such geometry features to `Circular Transmon Single Island`. If you are interested, you should then be able to perform simulations to calculate EPR of the qubit, even while varying the parameters to change the shape of the qubit.

## Motivation - how EPR is calculated using MER

The general goal of EPR is to design the geometry of the qubit with lowest TLS (Two-Level System) losses possible. TLS losses are proportional to the electric field energy and they are located mainly in the nanometer-scale oxidation layers (interfaces) of the device. EPR (Energy Participation Ratio) is the electric field energy in some part of the device divided by the total energy of the device. Thus, EPR is a measure how much energy is in that part and therefore how large TLS losses are expected. It is possible to simply calculate EPR for the entire 3D object. However, to have the accuracy of such simulation to produce useful results (in particular, in the interfaces), the simulation time would be unfeasibly long and would use up unfeasible amount of memory. However, accuracy is an issue at selected parts of the qubit, more specifically at the metal edges of the qubit (due to high gradients in the electric field), which is where metal deposit ends and exposure of underlying substrate layer begins. At such parts of qubit geometry, we are forced to model explicitly the oxidation layers, which contribute a lot to qubit's propensity to lose its energy state, resulting in low T1.

Now instead of 3D simulations, consider 2D simulations. Since geometry domain is so much more limited, we can model geometry, including oxidation layers on top of metal and substrate shapes, and spend fraction of the time running such simulations. Now, we identify key regions in qubit shape such that this region due to its shape can be characterized with a cross section, integrated over its perimeter. So what we do is we run a baseline heavy calculation over the entire 3D qubit with no oxidation layers defined (as geometries), then for each key region we run a lightweight 2D simulation for the cross section in 3D geometry. Cross-section simulation gives us an accurate result, which we integrate over the perimeter to get a result for the region, then we correct the result of the lower accuracy and heavy weight 3D simulation using the region results.

Generally the goal is to shift this ratio towards vacuum and substrate, and away from oxidized layers. It is also valuable information to the designer to know which regions in top-down qubit design have higher EPR.

## Code organization

We separate geometry definitions intended for EPR calculations from the [overall element design](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/qubits/circular_transmon_single_island.py). When we define EPR geometry, we create a python file of same name (`circular_transmon_single_island.py`) in [klayout_package/python/kqcircuits/simulations/epr](https://github.com/iqm-finland/KQCircuits/tree/main/klayout_package/python/kqcircuits/simulations/epr) folder. Feel free to check out other files in this folder for reference.

Let's have a closer look at python file [example.py](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/epr/example.py). This file does not define partition regions and correction cuts for any real element - its purpose is to give a simplified view of what is expected to be implemented in a real EPR python file. From this file we see what we need to implement:

- `partition_regions`: Function takes a [Simulation](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/simulation.py) object and returns a list of [PartitionRegion](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/partition_region.py) objects. The ordering of partition regions is important: If partition regions in the list overlap geometrically, the region later in the list will have its geometry substracted by all preceding partition regions. Thus all regions will be disjoint. The arguments to use for `PartitionRegion()` constructor:
  - `name` name of partition region as string. Please add "mer" suffix to the name. If you look up other examples in KQCircuits, you will see some PartitionRegions named with "bulk" suffix. For simplicity, let's only add "mer" partition regions for our qubit.
  - `region` can be [DBox](https://www.klayout.de/doc-qt5/code/class_DBox.html), [DPolygon](https://www.klayout.de/doc-qt5/code/class_DPolygon.html), list of those, or [Region](https://www.klayout.de/doc-qt5/code/class_Region.html). When simulation geometry gets exported, KQCircuits will detect metal edges within this `region` and cover only those parts. Can also be set to `None`, in which case the entire extent of the element is covered.
  - `metal_edge_dimensions`: List of two values, first value sets how far region expands on metal edge to the gap side, second value sets expansion to the metal side. If set to scalar value, that value will be used for both purposes. We recommend setting this to `2.0` for this issue for all partition regions.
  - `visualise` set to True for debugging reasons.
- `correction_cuts`: Function takes a [Simulation](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/simulation.py) object and returns a dictionary mapping a string to a dictionary. The dictionary must have at least `"p1"` and `"p2"` fields, that map to a [DPoint](https://www.klayout.de/doc-qt5/code/class_DPoint.html) object, effectively defining a line segment from `p1` to `p2`. The string key for the dictionary must match a `name` attribute of some `PartitionRegion` that is returned in `partition_regions` function. For one cut you will need to set `boundary_conditions` field, we will come back to this later in the description.
- `extract_from`: NOT NEEDED FOR UNITARYHACK. Implementing this allows to have composite simulations, for example coupling this qubit with another qubit, such that their corresponding partition regions are reused.

The use of [Simulation](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/simulation.py) object passed to `partition_regions` and `correction_cuts` functions should be limited to following:

- Read-only, don't modify the `Simulation` instance
- Usage of `refpoints` dictionary. Some key points in [CircularTransmonSingleIsland](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/qubits/circular_transmon_single_island.py) are stored as "refpoints" using `self.add_port()` calls, for example `"port_coupler_1"` and `"port_drive"`. Refpoints can also be stored directly using `self.refpoint["..."] = ...`. Some refpoints are added by default, like `"base"` refpoint at element's origin. You can reuse these points in the `klayout_package/python/kqcircuits/simulations/epr/circular_transmon_single_island.py` script via `simulation.refpoints`. You can view refpoints accessible for the qubit in KLayout by showing the `texts/refpoints` layer. You are allowed to add new refpoints to [design code](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/qubits/circular_transmon_single_island.py) if you need them.
- Usage of parameters defined as `Param()` objects. Such parameters are accessible in KLayout by double clicking the placed element and navigating to `PCell parameters`, and they can be modified live. For example, the value of parameter `r_island` can be accessed in `klayout_package/python/kqcircuits/simulations/epr/circular_transmon_single_island.py` via `simulation.r_island`. Note that elements in list parameters like `couplers_a` are treated as strings, so they need to be cast to float: `float(simulation.couplers_a[0])`.
- Other uses of [Simulation](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/simulation.py) object should be avoided.

Image
In above screenshot refpoints of qubit are visualised by showing the `refpoints` layer.

If implemented correctly, partition regions and correction cuts should cover correct regions of the qubit regardless what values were set to qubit's parameters, which allows sweeping certain design parameters of the qubit and calculating EPR for each sweep.

Debugging "stability" of EPR shapes is best done visually. Edit [gui_config.py](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/epr/gui_config.py) by adding a `"Circular Transmon Single Island"` entry, and give a list of partition regions you defined by their `name`. Now, you should be able to repeat the steps in  [First look](#first-look) for `"Circular Transmon Single Island"` to visualize partition regions and correction cuts you just defined. If KLayout gives an error for the qubit, try to follow the code line numbers it gives you, especially those present in `kqcircuits/simulations/epr/circular_transmon_single_island.py`, to fix any possible bugs. If the shapes become visible without errors, try changing some design parameters in qubit shape and see that partition regions and correction cuts still make sense. If you make a change in the EPR shape code, there is no need to restart KLayout to apply the changes, you can choose `KQCircuits > Reload Libraries` from top toolbar to apply recent changes in KQC code.

Since the element you are working on is a qubit, it has a SQUID placed by default. Generally, we replace the SQUID with empty pads for simulation purposes. Hence, when previewing the qubit in KLayout, it's best to set `junction_type` to `Sim` to more accurately preview partition regions.

## Partition region and correction cut recommendations

For the MER-corrected EPR simulation method to be accurate, the 2D correction simulations must accurately represent the region of the 3D model being corrected. This means that each geometrically distinct feature in the 3D geometry should have its own MER and a corresponding cross-section simulation defined. This can be achieved by using partition regions defined as KLayout polygons overlaid on the 3D geometry. In general, there is no optimal choice for partition regions and cuts, and comparing different “good” choices is very difficult without an accurate reference. However, simplicity is usually a high priority when designing EPR shapes. Simple regions are easier to debug, maintain, and are less prone to bugs when the qubit shape is adjusted. It is also important to consider which regions are of interest in terms of energy density—typically indicated by a narrow gap between signal and ground. In this case, for example, we are mostly interested in the gaps between the central island and couplers, whereas the narrow traces connected to the couplers are not relevant as long as the couplers are grounded.

There are no _exact_ requirements for this issue about how the partition regions and correction cuts should be defined, but we provide our recommendations below which fulfill issue's requirements. We are happy to review your own design of regions and cuts if they are different from our recommendation.

### Partition regions

Below are the listed PartitionRegions we recommend with an illustration. Regions should just be coarse shapes, inside which metal edge regions are detected. The region shouldn't be too "tight", meaning that the `PartitionRegion` extents should have extra space around the region we are interested in, at least by the value of `metal_edge_dimensions`.

1. `{n}cplrmer`, replacing `{n}` with `0`, `1`, `2` etc. A partition region for each coupler in the qubit. The number of couplers in the qubit is dynamically defined: if you add a fourth value for `couplers_a`, `couplers_b`, `couplers_angle`, `couplers_arc_amplitude` and `couplers_width`, this will create a fourth coupler in the qubit. It is acceptable if the submission only supports up to three couplers, meaning that fourth and beyond couplers don't need to have partition regions. But if only two or one coupler is defined, the qubit shouldn't crash. The shape of a partition region could be a "pizza slice" with a tip cut off near qubit center. The "slice" encompasses the coupler extents + extra space as described above.
2. `leadsmer`. A partition region for the junction "hall". This can be a rotated rectangle (hint: use [DCplxTrans](https://www.klayout.de/doc-qt5/code/class_DCplxTrans.html) to rotate shapes) of length `ground_gap`, width 8µm (junction hall width hardcoded in design), and angled at `squid_angle`. Remember that the shape should also have extra space around the "hall".
3. `bcomplementmer`. A partition region for main island. This can be defined as a complement region. Instead of defining a circle around the qubit, we can define `region=None`, which defines a region encompassing the entire qubit. If this `PartitionRegion` is the last one in the list, it will have all above partition regions substracted from it, leaving behind what is left, which is the island.

Remember to list these regions in [gui_config.py](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/epr/gui_config.py) for GUI debugging. Notice that `bcomplementmer` will not be visible in GUI since region is defined as `None`.

### Correction cuts

Correction cuts work best if they are perpendicular to any edge that they cross over. To guarantee this, its best to define cuts (when appropriate) to be segments of a line starting from qubit center extending outside at some direction.

1. `{n}cplrmer` crosses the coupler on some side of the coupler waveguide. Cut starts outside of the qubit, cuts through the coupler, then crosses into the qubit island and ends there. To have the cut stable wrt arbitrary `couplers_amplitude` values, the cut can be a line pointing to the center of the qubit angled at midangle between `couplers_angle[n]` and `couplers_amplitude[n]/2`. It should be avoided that the cut travels along the coupler waveguide.
2. `leadsmer` cut crosses the width of the junction hall. The cut should extend the junction hall extents to a "fair" amount, at least by `metal_edge_dimensions` value, as long as it doesn't cross over to another coupler or outside of qubit. In addition to `"p1"` and `"p2"` fields, for this cut there should also be a `"boundary_conditions"` field with following value: `{'xmin': {'potential': 0}, 'xmax': {'potential': 0}}`.
3. `bcomplementmer` cut crosses qubit's gap, that is it starts at island region and ends outside of qubit, extended on both ends at least by `metal_edge_dimensions` value. If the cut is defined at one direction, do notice that a coupler or a junction may be placed at the trajectory of this cut. It would be nice if in the cut definition this could be detected, and the cut placed somewhere else. Alternatively, you could document which direction in qubit is reserved for this cut, advising that for EPR simulations couplers and junctions shouldn't be placed at such directions.

Image

## Simulation script

> [!NOTE]
> To be eligible for the UnitaryHack bounty, you are not required to follow this subsection. You just need to include the simulation script as attached below in your pull request. You don't need to make sure that the script works in your machine, we can check the script for you on our devices. But we warmly recommend that you are able to see the simulation work for yourself.

A simulation script is a standalone script that imports geometries defined in KQCircuits and configures an external simulating software to correctly simulate such geometries. Such simulation scripts are listed in [klayout_package/python/scripts/simulations](https://github.com/iqm-finland/KQCircuits/tree/main/klayout_package/python/scripts/simulations) folder.

Let's take [circular_capacitor_epr_sim.py](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/scripts/simulations/circular_capacitor_epr_sim.py). If you run this python script from command line, it will prompt you with a KLayout window showing cross sections' geometry at correction cut locations. When you close that window, it will then prompt you with another KLayout window showing the circular capacitor design from top-down view. You can close that window too. Now, you should have a new folder `KQCircuits/tmp/circular_capacitor_epr_sim_output`. In it, there are a bunch of .gds files containing geometry data, paired with .json files that contain configuration for the external simulator, Elmer in this case. The folder also contains `epr.oas` and `simulation.oas`, which shows the same KLayout instance that were prompted previously after running the script. Each instance is a collection of multiple layouts, which can be viewed one at a time by picking a cell in left top `Cells` panel, right-clicking an entry and choosing `Show as new Top`. The `simulation.oas` file contains a sweep of different parameter values of the capacitor, while `epr.oas` contains those sweeps' cross sections at correction cuts.

Just running the python script does everything except for actually simulating. You can do this to do final checks that the simulations are properly exported, as running the actual simulation can take a long time. For example you can review `simulation.oas` for the following:

- Element shape features sweeped correctly. Pick `Show Only Selected` for the `1t1_gap` to simplify the design preview. Then pick `Show As New Top` for each cell in `Cells` list.
- Correction cuts placed correctly in each sweep, double-click the `cross_section_cut` layer at the bottom of layer listing
- Partition regions partitioned correctly. Double click layers starting with `vacuum` to inspect them. `vacuumbcomplement` layer should have previous partition regions carved out. Layers ending with `mer` should detect edges of the `1t1_gap` shape encompassed by the corresponding `bulk` layer.

Image

But we are simulating `Circular Transmon Single Island`, not `Circular Capacitor`. For this issue we provide a ready-made simulation script. Once you have `partition_regions` and `correction_cuts` implemented in `klayout_package/python/kqcircuits/simulations/epr/circular_transmon_single_island.py`, the script should work "out-of-the-box". Please include this script in your pull request at `klayout_package/python/scripts/simulations/circular_transmon_single_island_epr_sim.py`:

[circular_transmon_single_island_epr_sim.zip](https://github.com/user-attachments/files/20227047/circular_transmon_single_island_epr_sim.zip)

Finally, to actually run the simulation after exporting it, run this command from terminal:

`kqc sim circular_transmon_single_island_epr_sim.py`

It will prompt Gmsh window to showcase meshing of the simulation. After closing this window, it will perform the simulation (shouldn't take longer than 10 minutes on a standard laptop), after which it will prompt a Paraview window to visualize simulation results. It requires some manual steps to properly view the results in Paraview instance, check this unitaryhack issue for more details ([link](https://github.com/iqm-finland/KQCircuits/issues/108)). To view the results at a later time, run `paraview <PATH_TO_.pvtu_FILE>` in terminal.

You can disable all prompting by running simulation in "quiet" mode:

`kqc sim circular_transmon_single_island_epr_sim.py -q`

After simulation is done, all results are exported to `KQCircuits/tmp/circular_transmon_single_island_epr_sim_output`. EPR results are stored in `circular_transmon_single_island_epr_sim_output_epr.csv`. Columns starting with `p_` report the ratio of energy participation for each partition region for each layer.

If you get excited simulating this qubit, after submitting the pull request, here are some things you can try next:

- Sweep qubit shape to see which geometry feature improves the EPR. There is a commented block in the script that provides an example of sweeping two features. Be aware that this will increase the simulation time.
- Remove `ignore_ports`. This will cause the simulation to include waveguides being connected to the couplers of the qubit, and excite them under different signals. This would require adding partition regions and correction cuts for the new waveguides.

## Definition of done

- Define `partition_regions` and `correction_cuts` for `Circular Transmon Single Island` in `klayout_package/python/kqcircuits/simulations/epr/circular_transmon_single_island.py`
- Add `klayout_package/python/scripts/simulations/circular_transmon_single_island_epr_sim.py` script, can be unmodified from issue description. The validity of the script doesn't need to be tested by submitter.
- No unit tests or documentation update needed

AbdullahKazi500:
Files Created
1. klayout_package/python/kqcircuits/simulations/epr/circular_transmon_single_island.py
This file defines the partition regions and correction cuts needed for EPR simulation:
Partition Regions

Coupler Regions ({n}cplrmer)

One partition region for each coupler (dynamically created based on coupler parameters)
Shape: "Pizza slice" with tip cut off near qubit center
Encompasses each coupler with extra space around it
Uses metal_edge_dimensions=2.0 as recommended

Junction Hall Region (leadsmer)

Covers the Josephson junction "hall" area
Shape: Rotated rectangle positioned at squid_angle
Length: ground_gap + 10μm (with extra space)
Width: 8μm + 4μm (junction hall width + extra space)

Main Island Region (bcomplementmer)

Complement region covering the main island
Defined with region=None to cover entire element extent
Has all other partition regions subtracted from it (automatically handled by KQCircuits)

Correction Cuts

Coupler Cuts ({n}cplrmer)

Perpendicular cuts through each coupler
Start outside the qubit, cross through coupler, end at island edge
Angled at the coupler's angle for stability

Junction Hall Cut (leadsmer)

Crosses the width of the junction hall
Perpendicular to the hall direction
Includes required boundary conditions: {'xmin': {'potential': 0}, 'xmax': {'potential': 0}}

Main Island Cut (bcomplementmer)

Crosses from island to outside the qubit
Positioned to avoid conflicts with couplers and junction
Uses angle opposite to junction, with conflict detection and adjustment

Fixes #109


klayout_package/python/kqcircuits/simulations/epr/circular_transmon_single_island.py
```diff
@@ -0,0 +1,194 @@
+# This code is part of KQCircuits
+# Copyright (C) 2024 IQM Finland Oy
```
qpavsmi, line 2:
```suggestion
# Copyright (C) 2025 IQM Finland Oy
```

qpavsmi:
Please include this script in this branch at file location: `klayout_package/python/scripts/simulations/circular_transmon_single_island_epr_sim.py` - (https://github.com/user-attachments/files/20227047/circular_transmon_single_island_epr_sim.zip)

Also don't forget to do this part from issue description:

> Remember to list these regions in [gui_config.py](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/epr/gui_config.py) for GUI debugging. Notice that bcomplementmer will not be visible in GUI since region is defined as None.

qpavsmi:
> > > > Thanks! I will start looking through this starting tomorrow. Reminder that after today's deadline there will be a deadline on Monday for submissions that are in the review process.
> > > 
> > > 
> > > Hi it seems like there is another PR made to fix the issue
> > 
> > 
> > Yes, and in unitaryhack hackers are allowed to compete for the same issue. Whichever PR gets completed first, will get assigned the bounty
> 
> I hope you understood my concern about the other PR Thanks for understanding

The other PR is empty so I probably won't be accepting it at this stage. I don't think others are able to modify your branch, and if they are, they can only push commits which I will be able to differentiate. To be safe, you could make sure that you preserve your branch locally so if someone manages to overwrite your PR, you can overwrite it back.

qpavsmi:
image

For `bcomplementmer` correction cut, could you extend `bcomplementmer_1` point closer to qubit center, such that it is the same distance away from gap region as `bcomplementmer_2` point?

klayout_package/python/kqcircuits/simulations/epr/circular_transmon_single_island.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+def partition_regions(simulation):
+    """Returns list of PartitionRegion objects for Circular Transmon Single Island EPR simulation."""
+    
+    regions = []
+    
+    # Coupler partition regions - one for each coupler
+    for i, (c_angle, c_width, c_arc_ampl) in enumerate(
+        zip(simulation.couplers_angle, simulation.couplers_width, simulation.couplers_arc_amplitude)
```
qpavsmi, line 24:
```suggestion
    for i, (c_angle, c_arc_ampl) in enumerate(
        zip(simulation.couplers_angle, simulation.couplers_arc_amplitude)
```
It doesn't look like `c_width` is being used, so this for loop can be simplified

qpavsmi:
> > Also don't forget to do this part from issue description:
> > > Remember to list these regions in [gui_config.py](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/epr/gui_config.py) for GUI debugging. Notice that bcomplementmer will not be visible in GUI since region is defined as None.
> 
> Hi @qpavsmi I did not understood this part

In [gui_config.py](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/epr/gui_config.py) add following entry to `epr_gui_visualised_partition_regions` dict:

```python
    "Circular Transmon Single Island": [
        "0cplrmer",
        "1cplrmer",
        "2cplrmer",
        "leadsmer"
    ]
```

qpavsmi:
> and about the part for adding the script at the branch location can you confirm if I have added it

You've added it but it is modified from the file I attached. If you could use the contents of attached file unmodified

klayout_package/python/kqcircuits/simulations/epr/circular_transmon_single_island.py
```diff
@@ -0,0 +1,194 @@
...
+        # Create pizza slice polygon
+        points = [pya.DPoint(0, 0)]  # Center point
+        
+        # Add arc points for outer edge
+        n_points = 32
+        for j in range(n_points + 1):
+            angle = c_angle_rad - half_angle + (2 * half_angle * j) / n_points
+            x = outer_radius * math.cos(angle)
+            y = outer_radius * math.sin(angle)
+            points.append(pya.DPoint(x, y))
```
qpavsmi, line 47:
```suggestion
            points.append(pya.DPoint(x, y) + simulation.refpoints["base"])
```
The parititon regions should be shifted so their origin is at "base" refpoint of `CircularTransmonSingleIsland`. See also the following suggestions

klayout_package/python/kqcircuits/simulations/epr/circular_transmon_single_island.py
```diff
@@ -0,0 +1,194 @@
...
+        pya.DPoint(-hall_width/2, hall_length)
+    ])
+    
+    # Position at island edge and rotate
+    junction_pos = pya.DPoint(
+        simulation.r_island * math.cos(squid_angle_rad),
+        simulation.r_island * math.sin(squid_angle_rad)
+    )
+    
+    hall_transform = pya.DCplxTrans(1, simulation.squid_angle - 90, False, junction_pos)
```
qpavsmi, line 84:
```suggestion
    hall_transform = pya.DCplxTrans(1, simulation.squid_angle - 90, False, junction_pos + simulation.refpoints["base"])
```

klayout_package/python/kqcircuits/simulations/epr/circular_transmon_single_island.py
```diff
@@ -0,0 +1,194 @@
...
+    )
+    outer_point = pya.DPoint(
+        (simulation.r_island + simulation.ground_gap + 10) * math.cos(complement_angle),
+        (simulation.r_island + simulation.ground_gap + 10) * math.sin(complement_angle)
+    )
+    
+    cuts["bcomplementmer"] = {
+        "p1": inner_point,
+        "p2": outer_point
+    }
```
qpavsmi, line 192:
```suggestion
    cuts["bcomplementmer"] = {
        "p1": inner_point + simulation.refpoints["base"],
        "p2": outer_point + simulation.refpoints["base"]
    }
```

qpavsmi:
Finished with this round of review. I will summarize here what still needs to be done in prioritized order:

1. Fix the coupler correction cuts: https://github.com/iqm-finland/KQCircuits/pull/115#issuecomment-2965674812
2. Remove center point from partition regions: https://github.com/iqm-finland/KQCircuits/pull/115#discussion_r2142080059
3. `bcomplementmer` correction cut should be extended a bit towards the center of qubit: https://github.com/iqm-finland/KQCircuits/pull/115#issuecomment-2965638774
4. Add simulation script unmodified from the attached: https://github.com/user-attachments/files/20227047/circular_transmon_single_island_epr_sim.zip
5. Add an entry to `gui_config.py` https://github.com/iqm-finland/KQCircuits/pull/115#issuecomment-2965727020 This can benefit you too since you can then see the changes you make when working on 1. 2. and 3. on KLayout GUI
6. Remove unused `arc_points` import https://github.com/iqm-finland/KQCircuits/pull/115#discussion_r2142094767

With these done, I will check again if they were fixed correctly, and I will then assign you for the bounty. I will be limited in my ability to review during this weekend since I'll be traveling without my laptop. Following being my timezone GMT+2, I won't be available starting from Friday 16:00 until Sunday noon. I will be available on Monday for the entire day until the hard deadline lapses.

qpavsmi:
> BTW Where can we locate this gui. config Py in the repository ?

https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/epr/gui_config.py

qpavsmi:
> > > done now the only thing left is to add the script BTW is there any comments on the 1, 2, 3 and 6 reviews
> > 
> > 
> > All except 4. now work as proposed, thank you for the quick fix. Remember, you only need to include the script already supplied, no changes to the script is needed (I've already tested it works using the code in this pull request).
> 
> Thanks @qpavsmi BTW How to locate the script in the zip folder ? any specific name or such

its the only file

qpavsmi:
> > > done now the only thing left is to add the script BTW is there any comments on the 1, 2, 3 and 6 reviews
> > 
> > 
> > All except 4. now work as proposed, thank you for the quick fix. Remember, you only need to include the script already supplied, no changes to the script is needed (I've already tested it works using the code in this pull request).
> 
> And we should add it to the same path right ?

same path `klayout_package/python/scripts/simulations/circular_transmon_single_island_epr_sim.py`

PR #118, opened 2025-09-12
Add visualisation of simulation ports in KLayout preview

### LINKED ISSUE #52: Display ports in simulation export Klayout dialog
# Issue specification for UnitaryHack participants

## Prerequisites

KQCircuits, [GUI](https://iqm-finland.github.io/KQCircuits/developer/setup.html) and [standalone](https://iqm-finland.github.io/KQCircuits/developer/standalone.html) installations

## Issue description

When exporting files to be used by an external FEM simulation software, KQCircuits will usually show a preview of the geometry files that will be used for simulations in KLayout to review that the correct geometry is placed in correct layers. We also export some coordinates within geometry as "ports" in separate json files to be used by the simulation software. These ports do not show up in KLayout, and it's inconvenient to look them up from a separate file to see how the coordinates match with the design geometry. During the preview it would be good to have the port coordinates be outlined as text objects in KLayout.

## How to get started

Go to root of the KQCircuits source code folder and run:
`python klayout_package/python/scripts/simulations/xs1_full_chip_sim.py`

This will cause a KLayout window to pop up, showing geometry of an example chip for which a simulation can be run (remove some clutter double-clicking the `refpoints` layer to hide it). The python script will also generate the `tmp` subfolder if not present already, with a folder `tmp/xs1_full_chip_sim_output` inside. Inside that folder there is `simulation.oas` design file, which is exactly what the pop up window shows.

Review the contents of the `tmp/xs1_full_chip_sim_output/Simulation.json` file. Look through the `"ports"` list. You will see that ports numbered 1-8 are of type `EdgePort`, and the rest of the ports are of type `InternalPort`. These are the two classes of [ports](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/port.py) that are used in KQCircuits. EdgePorts are placed at the edge of the simulation region and InternalPorts are placed inside the region. Every port is defined by `signal_location` 2D coordinate, InternalPorts may additionally have a `ground_location` 2D coordinate.

You can compare the port locations as exported to `tmp/xs1_full_chip_sim_output/Simulation.json` with how we define the ports in [single_xmons_full_chip_sim.py](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/single_xmons_full_chip_sim.py). There are two calls of `self.ports.append`, `EdgePort`s are placed for the launchers and `InternalPort`s are added for the SQUID parts of qubits, signal locations located at the edge of qubits' inner metal part and ground locations located outside of qubits. See below charts.

edgeports

internalports

These would be good to illustrate by placing text objects like in `refpoints` layer but to a separate layer `simulation_ports`. Names of the text object should specify the port index and type, for example `edge_port_1`, `internal_port_signal_2`, `internal_port_ground_3`. However, `simulation_ports` layer should **not** be defined in [default_layer_config.py](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/layer_config/default_layer_config.py) like `refpoints` is, because simulation layers are also not defined there. Instead the `simulation_ports` layer should be used directly with `self.cell.shapes(self.layout.layer(layer)).insert(...)` where `layer` is a [LayerInfo](https://www.klayout.de/doc-qt5/code/class_LayerInfo.html) object of the `simulation_ports` layer. Value of `layer.layer` should be some unused number above 1000 (e.g. `len(simulation_layer_dict) + 1000` like in [simulation.py](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/simulation.py)), `layer.datatype` should be 0 and `layer.name` should be `"simulation_ports"`. The ports should be [DText](https://www.klayout.de/doc-qt5/code/class_DText.html) with `x` and `y` values set to port's coordinate and `string` set to port name. You can also play around with font size to distinguish the ports from refpoints.

Now try running:
`python klayout_package/python/scripts/simulations/double_pads_sim.py`

This script is different because rather than exporting one simulation, it sweeps simulations by exporting the same Double Pads qubit with varying coupler width. The folder `tmp/double_pads_sim_output_elmer` contains multiple json files, one for each Double Pads variation. Open `simulation.oas` file and you will see that each Double Pads variation is its own cell. You can view each of them separately by right clicking a cell entry and choosing `Show As New Top`. We want each such cell to contain only the ports that belong to that individual simulation sweep.

showasnewtop

The code in the [simulation.py](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/simulation.py) file should help the most with implementing this feature.

## Definition of done

- All ports exported to simulation json files should be visible in `simulation.oas` in a separate layer of text objects. Ports should convey their location, index and type of port as defined in the json files.
- The layer should not be statically defined in [default_layer_config.py](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/layer_config/default_layer_config.py) but be a dynamic layer like all simulation layers.
- If the simulation script creates a sweep of simulations (e.g. [double_pads_sim.py](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/scripts/simulations/double_pads_sim.py)), the ports should be separated to each appropriate simulation sweep
- The simulation GDS files should **not** include these port visualisations
- No tests needed

AbdullahKazi500:
Summary

This PR enhances the simulation preview in KLayout by adding visualisation for simulation ports.
Previously, ports (both EdgePort and InternalPort) were only exported to JSON files, which made it inconvenient to correlate their locations with the design geometry in the KLayout preview. With this change, ports are now displayed directly in KLayout on a dedicated layer (simulation_ports).

Changes

Added a new method Simulation.visualise_ports:

Iterates over all ports in self.ports.

Adds labels for each port with index and type (e.g., edge_port_1, internal_port_signal_2, internal_port_ground_2).

EdgePorts:

Visualised as text plus a short line segment extending outward from the simulation box edge.

InternalPorts:

Signal and ground locations are labelled separately.

A line is drawn between signal and ground locations when both are defined.

Called self.visualise_ports() at the end of Simulation.__init__, after geometry and ports have been built.

Ports now appear in the simulation preview OAS file (simulation.oas) in a dedicated layer simulation_ports.

fixes #52


qpavsmi:
> Hi Pavel any updates on this

Hello, apologies I couldn't look into this PR this week, I'll try to review it in the coming week. Also a coworker that is more knowledgeable on simulation matters is coming back form vacation next week so they can help to give more thorough review.

tmyllaeri:
Thanks for the PR @AbdullahKazi500 ! It seems that maybe you have forgotten to update the branch with the latest changes?

> Called self.visualise_ports() at the end of Simulation.init, after geometry and ports have been built.

This is missing and the indentation in the new function is incorrect. Otherwise looks great!

klayout_package/python/kqcircuits/simulations/simulation.py
```diff
@@ -1455,3 +1455,39 @@ def visualise_region(
         else:
             for i, p in enumerate(points):
                 self.cell.shapes(visualisation_layer).insert(pya.DText(f"{label}_{i+1}", p.x, p.y))
+   def visualise_ports(self):
```
qpavsmi, line 1458:
```suggestion
   def visualise_ports(self, thickness=50):
```
The port visualisation thickness could be passed as argument, but given some default value like 50 microns. Then this cab decide the thickness of the shape for both internalport and edgeport

qpavsmi (commented):
https://github.com/iqm-finland/KQCircuits/pull/118/files#diff-a86269e51682ac6034678eeff271f0c2e43d2e80ae19aa57ea37cba3a450ba4aR904

At the end of `create_simulation_layers`  function you could add the `visualise_ports` call.

klayout_package/python/kqcircuits/simulations/simulation.py
```diff
@@ -901,7 +901,87 @@ def create_simulation_layers(self):
...
+        """Visualise all ports in a dedicated 'simulation_ports' layer."""
+        dbu = self.layout.dbu
+        for port in self.ports:
+            if isinstance(port, EdgePort):
+                label = f"edge_port_{port.number}"
+                p = port.signal_location
+
+                # Triangle parameters
+                extension = 50   # arrow length in µm
+                thickness = 10   # base width in µm
```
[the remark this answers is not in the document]
qpavsmi: Maybe extension=150 and thickness=100 would be better defaults.

qpavsmi:
Apologies for changing requirements while working on this issue half-way through. I had a meeting with my team member and they brought up that we already define polygon shapes for ports. They can be accessed using `get_port_data()`, which returns a list  of dicts, each representing a port in simulation. It has "polygon" ``(get_port_data()[0]["polygon"]``) field, which is a list of 3D points. If you ignore the third coordinate, you could construct a 2D polygon which you can draw in "simulation_ports" layer.

This should work well for internal ports. For edgeports it will only draw a 0 thickness line, since edgeports are drawn as rectangles along 3D space. So maybe for edgeports it can be a line that is thickened in the direction away from simulation box, maybe something like this:

image

Try for yourself and look at values that `get_port_data()` returns, See what solution makes most sense to you. I've been reviewing your PR by running `python klayout_package/python/scripts/simulations/xs1_full_chip_sim.py` since that has nice combination of edgeports and internal ports.

I guess for internal ports that only have signal port but no edgeport there is no reason to draw any polygon, unless `get_port_data()` returns something. My team member was of the opinion that we never define ports this way I guess.

qpavsmi:
Anyway my workday is ending soon and I'll be off work until Monday, so I probably won't be able to review your updates swiftly.

qpavsmi:
> oh but I still did not understand on how to get the values

First get the port data as json with:

```
port_json = self.get_port_data()
```

Then when you iterate over ports, find a port entry from the json by current ports number:

```
port_data = [p for p in port_json if p["number"] == port.number][0]
```
This will crash if `port_json` does not have an entry with `"number"` field that equals `port.number`, so best to check for this and print warning if that's the case, then don't draw this port.

Lastly, construct the polygon:

```
poly = pya.DPolygon([pya.DPoint(p[0], p[1]) for p in port_data["polygon"]])
```

then you can create a region out of the polygon like you already do in your code:

```
region = pya.Region(poly.to_itype(dbu))
```

qpavsmi:
> Hi Pavel Since get_port_data() already returns polygons for ports, we can directly use those for visualisation instead of re-inventing shapes. For InternalPorts, this polygon is a proper area, while for EdgePorts it’s a thin rectangle (0 thickness in 2D), so we can decide whether to thicken or extend outward.

That's correct. Let's extend the thin rectangles of EdgePorts outwards from the simulation box.

Apologies for leading you astray in your previous efforts, I was reminded by my coworker later on that we already generate polygons for ports.

klayout_package/python/kqcircuits/simulations/simulation.py
```diff
@@ -1455,3 +1454,54 @@ def visualise_region(
...
+            logging.warning(f"Unsupported port type {type(port).__name__} for port {port.number}")
+            continue
+
+        # Construct 2D polygon from port_data["polygon"] (ignoring z coordinate)
+        if "polygon" in port_data and port_data["polygon"]:
+            points_2d = [pya.DPoint(p[0], p[1]) for p in port_data["polygon"]]
+
+            if isinstance(port, EdgePort) and edge_port_thickness > 0:
+                # Extend outward depending on which edge the port is on
+                poly_box = pya.DBox().around(points_2d[0], points_2d[1]) if len(points_2d) >= 2 else None
```
qpavsmi, line 1489:
```
                poly_box = pya.DBox().around(points_2d[0], points_2d[1]) if len(points_2d) >= 2 else None
```
Confused about this line. `around` doesn't seem to be a function defined for `DBox`. Also `poly_box` is not used anywhere else in your code.

klayout_package/python/kqcircuits/simulations/simulation.py
```diff
@@ -1455,3 +1457,63 @@ def visualise_region(
...
+            label = f"internal_port_{port.number}"
+        else:
+            logging.warning(f"Unsupported port type {type(port).__name__} for port {port.number}")
+            continue
+
+        # Construct 2D polygon from port_data["polygon"] (ignoring z coordinate)
+        if "polygon" in port_data and port_data["polygon"]:
+            points_2d = [pya.DPoint(p[0], p[1]) for p in port_data["polygon"]]
+
+            if isinstance(port, EdgePort) and edge_port_thickness > 0 and len(points_2d) >= 2:
```
qpavsmi, line 1490:
```suggestion
            if isinstance(port, EdgePort) and edge_port_thickness > 0:
```
IMO `len(points_2d) >= 2` check doesn't really do anything, unless you disagree

klayout_package/python/kqcircuits/simulations/simulation.py
```diff
@@ -1455,3 +1457,73 @@ def visualise_region(
...
+                                p1 + direction,
+                            ]
+                        )
+                        visualise_point = port.signal_location
+            else:
+                # Internal ports → use polygon directly
+                poly = pya.DPolygon(points_2d)
+
+            region = pya.Region(poly.to_itype(dbu))
+            self.visualise_region(region, label, "simulation_ports")
```
qpavsmi, line 1527:
```suggestion
            self.visualise_region(region, label, "simulation_ports", visualise_point)
```

You forgot to add `visualise_point` argument. Also handle `visualise_point` definition if the port is not an edgeport.

Please fix the indentation now, otherwise the code cannot be run. Use `black -l 120 -t py310 .` to fix formatting in the file. Then we can merge this PR

qpavsmi:
- `visualise_ports` function still not indented correctly, code cannot be run
- Please add an empty line between `visualise_region` and `visualise_ports`, see other functions in `simulation.py` for reference
- `visualise_point` value is only set if port type is EdgePoint, it should have some value in all cases. Consider setting its value to None in other cases.
- `simulation.py` file name was changed to `simulation_formatted.py`, please change it back

klayout_package/python/kqcircuits/simulations/simulation.py
```diff
@@ -1455,3 +1457,74 @@ def visualise_region(
...
+        else:
+            logging.warning(f"Unsupported port type {type(port).__name__} for port {port.number}")
+            continue
+
+        # Construct 2D polygon from port_data["polygon"] (ignoring z coordinate)
+        if "polygon" in port_data and port_data["polygon"]:
+            points_2d = [pya.DPoint(p[0], p[1]) for p in port_data["polygon"]]
+
+            if isinstance(port, EdgePort) and edge_port_thickness > 0:
+                    direction = None
```
qpavsmi, line 1492:
```suggestion
            if isinstance(port, EdgePort) and edge_port_thickness > 0:
                direction = None
```
The code now I think is runnable (can't check because I'm running out of time for today) but indentation can still be improved for style. Try running `black -l 120 -t py310 klayout_package/python/kqcircuits/simulations/simulation.py` and it will fix the style automatically.

Also if you're unsure that the git tool pushed the changes you made correctly, refer to https://github.com/iqm-finland/KQCircuits/pull/118/files to see what is the code that I see.

klayout_package/python/kqcircuits/simulations/simulation.py
```diff
@@ -902,6 +902,8 @@ def create_simulation_layers(self):
             if part.visualise:
                 self.visualise_region(part.region, part.name, f"part_reg_{part.name}")
 
+        # Finally, visualise all ports
+        self.visualise_ports()
     def produce_layers(self, parts):
```
qpavsmi, line 907:
```suggestion
        self.visualise_ports()

    def produce_layers(self, parts):
```
Add empty line here please

klayout_package/python/kqcircuits/simulations/simulation.py
```diff
@@ -1455,3 +1458,81 @@ def visualise_region(
...
+
+            # Label based on port type
+            if isinstance(port, EdgePort):
+                label = f"edge_port_{port.number}"
+            elif isinstance(port, InternalPort):
+                label = f"internal_port_{port.number}"
+            else:
+                logging.warning(
+                    f"Unsupported port type {type(port).__name__} for port {port.number}"
+                )
```
qpavsmi, line 1490:
```suggestion
                logging.warning(f"Unsupported port type {type(port).__name__} for port {port.number}")
```
Sorry for having these endless tweaks, I promise that this set of suggestions is the last one :D Running the black tool wants this change, just click apply suggestion

klayout_package/python/kqcircuits/simulations/simulation.py
```diff
@@ -1455,3 +1458,81 @@ def visualise_region(
...
+                        visualise_point = port.signal_location
+                else:
+                    poly = pya.DPolygon(points_2d)
+
+                region = pya.Region(poly.to_itype(dbu))
+                self.visualise_region(region, label, "simulation_ports", visualise_point)
+            else:
+                logging.warning(
+                    f"Port {port.number} has no polygon data in get_port_data(), skipping visualisation."
+                )
```
qpavsmi, line 1538:
```suggestion
                logging.warning(f"Port {port.number} has no polygon data in get_port_data(), skipping visualisation.")
```

qpavsmi (approved):
Straight forward implementation to visualize ports defined for the exported simulation within the KLayout interface. Thanks for sticking through the development process!

qpavsmi (approved):

PR #130, opened 2026-06-03
Create test.py

### LINKED ISSUE #122: Export layout present in KLayout GUI as an Elmer simulation

---

### Description

Currently simulations can be defined for simulations that are defined through Python code, but there is no way to take designs that were manually drawn in KLayout GUI and extract the simulation from there. The task is to make a KLayout GUI macro that takes whatever geometry is in the layout, detect relevant elements and export to Elmer simulation which can be run.

### Prerequisites

> [!IMPORTANT]
> Completing this issue requires features from Elmer software that is at this time only available in Linux distribution. Windows developers can get this to work by installing a Linux image in Windows Sybsystem in Linux (WSL), see [here](https://learn.microsoft.com/en-us/windows/wsl/install). Unfortunately there is limited Elmer support for mac developers. If Elmer cannot be made to work on developer's device, this issue can be worked up to the point until Elmer is executed, and reviewers can check that Elmer-side features work correctly.

Confirm that KLayout application can be launched, then perform [KQC GUI installation](https://iqm-finland.github.io/KQCircuits/developer/setup.html). If using WSL, we recommend cloning KQCircuits to WSL file system. But it should be possible to mount windows directory in WSL. The command to install KQC for KLayout GUI usage is:

After restarting the KLayout application, you should see the `KQCircuits` entry in the top toolbar, and in `Libraries` tab at bottom-left you should be able to select entries like `Qubit Library` etc.

Next install [Elmer FEM simulator](https://www.elmerfem.org/blog/). Best way is to install it from singularity container. Guide available [here](https://iqm-finland.github.io/KQCircuits/developer/containers.html#singularity-usage). Windows users should install Elmer using WSL.

Currently available simulations in KQC require a standalone installation to work. It would be good to set this installation up to compare existing simulations with the one being developed for this issue. The standalone installation guide is [here](https://iqm-finland.github.io/KQCircuits/developer/standalone.html). We recommend ``Reproducible, Secure Installation`` option, command being:

```
cd klayout_package/python
python -m pip install -r requirements/<platform>/requirements.txt
python -m pip install -r requirements/<platform>/sim-requirements.txt
python -m pip install --no-deps -e .
```

### Use case

There are resources to catch you up-to-speed on how to use KLayout GUI to get a simple design done. See KQC docs on [placing elements](https://iqm-finland.github.io/KQCircuits/getting_started/gui_features/gui_elements.html) and [routing waveguides](https://iqm-finland.github.io/KQCircuits/getting_started/gui_features/gui_waveguides.html).

However, to save your time working on thie task, we provide a ready-made design we drew in GUI that you can use: a waveguide connecting two launchers that branches into two different qubit designs. Below is a .zip file containing a .oas file that can be opened with KLayout.
[gui_design.zip](https://github.com/user-attachments/files/28065972/gui_design.zip)

Image

Once .oas file is opened, from top toolbar choose `Load layer properties` and choose [the default_layer_props.lyp file](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/layer_config/default_layer_props.lyp) to have a cleaner view of the design.

The task is to take whatever layout is present, and to convert that into a simulation that can be run with Elmer. Let's look at an existing python script that exports a simulation as a reference for what to do for this task. Take [xs1_full_chip_sim.py](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/scripts/simulations/xs1_full_chip_sim.py). Executing this script will export simulation files:

```
python klayout_package/python/scripts/simulations/xs1_full_chip_sim.py
```

As a result, it will create a folder `KQCircuits/tmp/xs1_full_chip_sim_elmer` with some files, and prompt a KLayout window that opens file `KQCircuits/tmp/xs1_full_chip_sim_elmer/simulation.oas`. In the folder there is also `simulation.sh`/`simulation.bat` file that initiates Elmer to perform simulation on exported files. Warning that this can take a long time on a standard laptop for this particular simulation. Some simulations may have mutiple "sweeps", however this simulation only contains one "sweep" called `xs1_full_chip_sim`. For each sweep following files get created:
- The `xs1_full_chip_sim.gds` file is a geometry file similar to `simulation.oas` which contains only simulation related layers. This file has no exclusive data to `simulation.oas`, but this file is what Elmer will read.
- `xs1_full_chip_sim.json` file contains some parameters that the simulator can use, most notably the locations of the simulation ports.
- `xs1_full_chip_sim.sh`/`.bat` file is an executable that performs simulation steps, such as meshing with Gmsh, separate Elmer steps, then previewing using Paraview. This file is invoked from `simulation.sh`/`.bat`.

Let's have a closer look on `simulation.oas`:

Image

Displaying all layers is a bit cluttered, so best to show a few layers at a time:
- `1t1_gap` shows the "gap" regions, that is where metal is etched away to expose silicon. Geometry at this layer map to the original geometry from which simulation was exported from at layer `1t1_base_metal_gap_wo_grid`
- `substrate_1` and `vacuum` in this case is drawn where the simulation [box](https://github.com/iqm-finland/KQCircuits/blob/14483b00d20a4033971d08b99428b4ec9ace9d58/klayout_package/python/kqcircuits/simulations/single_xmons_full_chip_sim.py#L76) is defined. Simulation will not be performed outside of this box.
- `simulation_ports` previews where simulation ports are placed. `EdgePort`s are placed at the launchers, `InternalPort`s are placed where qubit junctions would be.
- `1t1_signal_*` layers have metal regions that are "activated" by the simulation ports. 1-7 signal layers are activated by launcher EdgePorts, 8-13 signal layers are activated by qubit juntion InternalPorts
- `1t1_ground` layer contains all other metal layers, which is grounded.

### Technical requirements

The macro you implement should produce a folder in `KQCircuits/tmp` directory with similar content to what `xs1_full_chip_sim.py` produces. The macro should automatically determine the "box" of the simulation, find [Launcher](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/elements/launcher.py) elements in the layout to place [EdgePorts](https://github.com/iqm-finland/KQCircuits/blob/14483b00d20a4033971d08b99428b4ec9ace9d58/klayout_package/python/kqcircuits/simulations/port.py#L120) there, then find qubits in the layout, set their `junction_type` to `Sim` and place [InternalPorts](https://github.com/iqm-finland/KQCircuits/blob/14483b00d20a4033971d08b99428b4ec9ace9d58/klayout_package/python/kqcircuits/simulations/port.py#L68) there.

Let's study how a simulation is exported in existing use case, let's take the [xs1_full_chip_sim.py script](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/scripts/simulations/xs1_full_chip_sim.py). Following is the core function call that exports the simulation files:
```
export_elmer(simulations, path=dir_path, workflow=workflow)
```
with passed arguments defined earlier in the script. `dir_path` is a file path to the folder that gets created for this simulation. `workflow` is a dictionary that configures the simulation execution pipeline. These should be simple enough to understand from this example, but let's delve deeper into how the `simulations` argument is constructed. The line in the script is 
```simulations = [(SimClass(pya.Layout(), **sim_parameters), solution)]```
Here, `SimClass` is aliased to [SingleXmonsFullChipSim](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/single_xmons_full_chip_sim.py), which during initialisation takes a [Layout](https://www.klayout.de/doc-qt5/code/class_Layout.html) object where the simulation geometry gets drawn, and a dict of `sim_parameters` that adjusts certain geometry features of the `SimClass`. Some simulation scripts intend to "sweep" simulation parameters to multiple values, hence `simulations` may be a list with multiple values. Since our macro will take static geometry in the layout, parametrizing geometry is not relevant for us, so we can work with singleton lists as `simulations`.

#### Simulation configuration

The [ElmerVectorHelmholtzSolution](https://github.com/iqm-finland/KQCircuits/blob/14483b00d20a4033971d08b99428b4ec9ace9d58/klayout_package/python/kqcircuits/simulations/export/elmer/elmer_solution.py#L136) object instructs Elmer to solve a wave-equation in the geometry to extract a so-called "S-matrix". This is an Elmer simulation type that is the best fit for simulating something that is quickly put together in a KLayout GUI session. The `Solution` object is a `dataclass` containing several configuration including the target `frequency` that a user can decide freely. The `mesh_size` determines the fidelity of the simulation: coarser mesh is faster to simulate but produces less accurate results.

Your macro should have a clear configuration section that the macro user can easily reconfigure if they wish to do so. For example macro user might decide to use a different `target_frequency`, or to set a different `mesh_size`. Following are recommended default configurations:

```
mesh_size = {
    "global_max": 200.0,
    **refine_metal_edges(10, 1.0),
}

solution = ElmerVectorHelmholtzSolution(
    mesh_size=mesh_size,
    frequency=5.0,
    # Use first order basis since the model is very large
    quadratic_approximation=False,
    linear_system_method="zmumps",
    use_multigrid_solver=False,
)

workflow = {
    "run_gmsh_gui": True,
    "run_elmergrid": True,
    "run_elmer": True,
    "run_paraview": True,
    "python_executable": "python",
    "gmsh_n_threads": -1,  #  Number of omp threads in gmsh
    "elmer_n_processes": -1,  # Number of dependent processes (tasks) in elmer
    "elmer_n_threads": 1,  # Number of omp threads per process in elmer
}
```

Now we just need to understand how a "Simulation class" such as [SingleXmonsFullChipSim](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/single_xmons_full_chip_sim.py) draws geometry and places the ports. In the simulation class code, the `build` function boils down to placing the [SingleXmons](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/chips/single_xmons.py) chip, deleting unneeded parts like chip frame and "test pads" (in our macro we won't need to remove any elements), sets the box with `self.box` assignment, then placing `EdgePort`s for launchers and `InternalPort`s for qubits. Basically your task is to write a macro that takes the geometry as drawn in KLayout session instead of drawing geometry using `build`, and to automatically determine the `box` and `ports` attributes based on elements present in the layout.

#### Instance hierarchy scraper

You should implement a scraper of the instance hierarchy of the active layout that finds following elements:
- [Launchers](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/elements/launcher.py)
- Qubit subclass elements, see [Qubit superclass](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/qubits/qubit.py) and [its subclasses](https://github.com/iqm-finland/KQCircuits/tree/main/klayout_package/python/kqcircuits/qubits)

Then the macro places the ports based on elements' refpoints. The scraping algorithm could resemble something like the [extract_junctions](https://github.com/iqm-finland/KQCircuits/blob/14483b00d20a4033971d08b99428b4ec9ace9d58/klayout_package/python/kqcircuits/util/replace_junctions.py#L219) function. The idea is to start with `top_cell` which in the macro can be obtained by:

```
view = KLayoutView(current=True)
top_cell = view.active_cell
```
[Cell](https://www.klayout.de/doc-qt5/code/class_Cell.html) is an abstract unit of geometry, and there may be multiple instances of cell placed in the layout. Cell's are agnostic in their position and orientation, but an [Instance](https://www.klayout.de/doc-qt5/code/class_Instance.html) of the cell also has a transformation [Trans](https://www.klayout.de/doc-qt5/code/class_Trans.html) which tells at what location to place cell content and how to orient (rotate, mirror) it. To iterate over child instances of the cell:
```
for inst in cell.each_inst():
```
Then to look up from which cell the given instance comes from:
```
layout = view.layout
cell = layout.cell(inst.cell_index)
```
To check if the instance is of some KQC class, for example launcher:
```
pcell = inst.pcell_declaration()
if isinstance(pcell, Launcher):
```
To get the transformation of Instance:
```
# Returns DCplxTrans type, see KLayout docs for details
inst.dcplx_trans
```
You will need to compound the transformations within hierarchy path starting from the top cell using multiplication operator:
```
trans = prev_trans * inst.dcplx_trans
```
To get the refpoints of a cell:
```
refpoints = get_refpoints(layout.layer(default_layers["refpoints"]), cell)
refpoints["base"]
```
"base" refpoint is placed at 0,0 coordinate of every KQC element. If you read class code of some `Element` subclass (like [Launcher](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/elements/launcher.py)), calling `self.refpoints["abc"] = pya.DPoint(5,6)` in `build` function will place that point within element coordinate system at given refpoint string.

To get a PCell parameter value, for example `junction_type` of qubits:
```
junction_type = inst.pcell_parameter("junction_type")
# or get all pcell parameters as dict
parameters = inst.pcell_parameters_by_name()
junction_type = parameters["junction_type"]
```
Seems to also have a way to set a PCell parameter:
```
inst["junction_type"] = "Sim"
```
If this works, this would be good to set for all qubits in the layout since our current simulation workflows cannot properly simulate well-defined junctions.

The part of the code in [SingleXmonsFullChipSim](https://github.com/iqm-finland/KQCircuits/blob/14483b00d20a4033971d08b99428b4ec9ace9d58/klayout_package/python/kqcircuits/simulations/single_xmons_full_chip_sim.py#L96) that places `InternalPort`s at qubit locations can be simplified for the macro to not use `etched_line` function, but to specify `signal_location` and `ground_location` directly via refpoints:
```
refpoints = get_refpoints(layout.layer(default_layers["refpoints"]), cell)
ports.append(
    InternalPorts(
        number=i,
        signal_location=trans * refpoints["squid_port_squid_a"],
        ground_location=trans * refpoints["squid_port_squid_b"]),
)
```

#### Simulation box and EdgePorts

The simulation box should be automatically obtained. A good starting point is to get the bounding box of all geometry within the cell:
```
box = top_cell.dbbox()
```
However, the `Edgeport`s for valid simulation can only be placed at the "rim" of the box. Hence the edges of the box should be adjusted such that the `Edgeport`s are placed on the edge. Launchers, when placed manually, may not align very accurately, so would be nice to have an algorithm that finds the box edges as illustrated in below picture:

Image

The `Edgeport`s ideally would be placed at the opposite edge of the launcher opening, but having the port closer to the opening is also OK. However macro should report an error if a box cannot be defined given the constraints of Launcher positions:

Image

With these, you can use the following function to convert the top cell into a [Simulation](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/simulation.py) instance:
```
simulation = Simulation.from_cell(top_cell, margin=100, name=name, box=box, ports=ports)
```
The [from_cell](https://github.com/iqm-finland/KQCircuits/blob/14483b00d20a4033971d08b99428b4ec9ace9d58/klayout_package/python/kqcircuits/simulations/simulation.py#L332) function is otherwise ready but it doesn't accept the `ports` argument in current codebase. Make sure to extend the function signature. You will also need to change the `Simulation.__init__` constructor signature so that `ports: list[Port]` can optionally be passed via `from_cell` function, setting internal attribute `self.ports` to that value. Reason is that once `Simulation` instance is constructed, it will immediately call `build` abstract function which is assumed to take care of the ports, and passing `ports` to already constructed `Simulation` object does nothing useful.

Now hopefully you should have everything you need to have the following function call at the end of the macro generate simulation files ready to be executed:
```
export_elmer([(simulation, solution)], path, workflow=workflow)
```

#### Previewing simulation

The `from_cell` function causes extra geometry to be generated, mainly the `1t1_signal_*` layers that partitions the metal regions to what are induced to some port's signal, and `1t1_ground` which contains the grounded metal. Ideally, executing the macro shouldn't cause modifications to the layout that the user designed. So it would be a good idea to have the macro open a new layout within KLayout session that creates above layers, while keeping the original layout in previous tab intact. You can cause KLayout to open new tab via:
```
view = KLayoutView()
```
after which you can copy `top_cell` to `view.layout` and call `simulation = Simulation.from_cell(...)` on the new layout. As a result, running the macro will open a new tab in KLayout which has the same design but with simulation layers, but original tab from which the macro was run should still be intact, with no changes caused to it. This behaviour is similar to the `open_with_klayout_or_default_application` call.

### Executing simulations

The responsibity of the macro you are tasked to implement is to only export the simulation to some folder, by default to `KQCircuits/tmp` folder. To verify that the simulation files your macro exports can be executed correctly, try running the `simulation.sh` script that gets exported to the folder. It should go through following stages in default configuration:
- Perform meshing using Gmsh, open meshing preview
- Execute Elmer, resulting in files
  - `*_project_results.json`
  - `*.s2p`
  - `<SIMNAME>/resonant_vtus/*.pvtu` files
- Run Paraview to open `*pvtu` files for visualisation

Try to look into these files, and feel free to ask us if your files look correct. Let us know if you used our example design [gui_design.zip](https://github.com/user-attachments/files/28065972/gui_design.zip), or if you drew your own design, make sure to save that as `.oas` file. You can then share the `.oas` file with us and we will export the simulation using the macro in your pull request, so we can verify that the simulation can be executed correctly.

If Elmer works on your device, you can also share with us the simulation folder after simulation was executed for us to review. Remember that for the purposes of this task, you are not required to have Elmer deployment working on your device, all we ask is that your implementation performs exporting correctly.

### Acceptance criteria

- New macro `klayout_package/python/scripts/macros/export/export_elmer.lym` takes geometry present in open layout and exports wave-equation simulation of that geometry to `KQCircuits/tmp` (or other configurable path)
- No unit tests required

l997RaghavAnand:
creating test .py fix #122

### DIFF
--- a/klayout_package/python/test.py
+++ b/klayout_package/python/test.py
@@ -1,0 +1,1 @@
+testing


qpavsmi:
Thank you for your submission and participation in unitaryHack. Unfortunately we decided to move forward with another submission

PR #131, opened 2026-06-03
Create newtest.py

### LINKED ISSUE #125: Migrate EPR shapes preview in GUI to pya.Marker

---

### Description

We have a feature to preview single element EPR shapes (partition regions and correction cross section cuts) on an active KLayout GUI instance. Currently the feature is implemented by specifying which existing layers to use to place the EPR shapes there to preview. This is problematic because the purpose of placing shapes to specific layer is to fabricate these shapes using some specified process. However EPR shapes are not fabricated, instead they instruct a FEM simulator to treat the region differently during simulation. Relying on existing layers temporarily risks actual design shapes "colliding" with EPR shapes and otherwise risks miscommunication within the layout.

It would be better to draw these shapes into a separate utlity that is completely separate from fabrication. For that we could use KLayout API's [Marker](https://www.klayout.de/doc-qt5/code/class_Marker.html) class.

### Prerequisites

Confirm that KLayout application can be launched, then perform [KQC GUI installation](https://iqm-finland.github.io/KQCircuits/developer/setup.html) with command:

After restarting the KLayout application, you should see the `KQCircuits` entry in the top toolbar, and in `Libraries` tab at bottom-left you should be able to select entries like `Chip Library` etc.

### Use case

Open KLayout GUI. Make sure in top tool bar `File > Setup` the `Application > Editing mode` has the checkmark on. Restart KLayout if you needed to set it on. Then choose `Qubit Library` in bottom left `Libraries`, drag the `Swissmon` into the layout, and click once. Press Esc to deactivate placing swissmon copies, press F2 to zoom out layout, and press star key `*` to make entire cell hierarchy visible. You should have Swissmon qubit in full view.

Double-click the swissmon, you should see new window `Instance properties` appear. See below screenshot with relevant parts outlined:

Image

Make sure to open the `PCell parameters` tab open. From `Options` enable `Show parameter names` to view parameter names as defined in KQC code.

All parameters prefixed with `_epr_` will allow you to preview the shapes defined in the EPR code. The `_epr_show` needs to be set to True for anything to be previewed. To view cross section cuts, a layer needs to be selected for `_epr_cross_section_cut_layer`. You can choose one of the top layers it will offer randomly to preview the cuts. Then for each partition region you can also choose separate layers to preview the partition regions using `_epr_part_reg_xxx_layer`.

You can now preview the cross section cuts and parititon regions of Swissmon, and you can also adjust other parameters that affect the design shape (for example `cpl_length`) and see that partition regions and cuts get updated to align with the qubit design.

### Background

The distinction between elements design shapes and EPR shapes is made clear in the code. The EPR shape definitions are put into dedicated folder [kqcircuits/simulations/epr](https://github.com/iqm-finland/KQCircuits/tree/main/klayout_package/python/kqcircuits/simulations/epr). In the case of swissmon, the design is implemented [here](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/qubits/swissmon.py) but EPR shapes are defined [here](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/epr/swissmon.py).

The EPR code has `partition_regions` function that returns the ordered `PartitionRegion` list. The code also has `correction_cuts` function that returns a dict of line segments that are contained within their respective PartitionRegion. These two functions take as input element instance's parameters that were initialized with `Param(...)` (e.g. `simulation.arm_length[0]`) and refpoints (e.g. `simulation.refpoints["epr_cross_06"]`).

Some key points about this GUI feature being implemented in the [Element](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/elements/element.py) code. Some EPR related parameters are defined the same way as other parameters are defined using `Param(....)`. However since a parameter is needed for each partition region, number of which varies for each element, these parameters are initiated dynamically at the tail-end of the [init function](https://github.com/iqm-finland/KQCircuits/blob/0ac500068dc7af1480c59758843dadbf49113e5c/klayout_package/python/kqcircuits/elements/element.py#L190). It is manually selected which parition regions we wish to be previewable in GUI, listed in [epr/gui_config.py](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/epr/gui_config.py). All correction cuts are visualised.

The `post_build()` [function](https://github.com/iqm-finland/KQCircuits/blob/0ac500068dc7af1480c59758843dadbf49113e5c/klayout_package/python/kqcircuits/elements/element.py#L486) is called by KLayout after calling `build()`. It has following two lines, which causes the cross section cuts and partition regions to show in GUI:

```
        self._show_epr_cross_section_cuts()
        self._show_epr_partition_regions()
```

These functions look up a module that share the same name as the `Element` subclass module, looks it up from [simulations/epr](https://github.com/iqm-finland/KQCircuits/tree/main/klayout_package/python/kqcircuits/simulations/epr) folder, imports the module dynamically and calls needed function, passing `self` as argument, which is expected to have appropriate `Param` parameters and refpoints to get correction cuts and partition regions.

### Implementation recommendation

All EPR related parameters of type `pdt.TypeLayer` should be changed to `pdt.TypeBoolean`, and if parameter is set to True, it will draw needed shape using ``Marker``s. `Marker`s can be defined using multiple shape types, including [DPolygons](https://www.klayout.de/doc-qt5/code/class_DPolygon.html) using Marker's `set_polygon` function, and [DText](https://www.klayout.de/doc-qt5/code/class_DText.html) using `set_text`. You can then customize Marker object visual attributes like `color`, `dither_pattern`, `vertex_size`, `line_width`, `halo` etc. We recommend that the polygons are filled with some dithering to see more clearly what regions the shape covers.

`Markers` are set on the [LayoutView](https://www.klayout.de/doc-qt5/code/class_LayoutView.html) object, using the `add_marker` function. Recommended way to access the active `LayoutView` object from [Element](https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/elements/element.py) code is by following:

```
from kqcircuits.pya_resolver import lay
layout_view = lay.LayoutView.current()
layout_view.add_marker(marker)
```
To clear all markers drawn so far:
```
layout_view.clear_markers()
```

There seems to be a way to clear individual markers, but it relies on marker variable going out-of-scope, which I'm not sure is reliable enough, see `Marker` docs. We would recommend clearing the layout every time a change in marker shapes is imminent, then redrawing them.

Part of this task is to implement this feature in such a way that a casual use of KLayout won't cause markers to linger when they shouldn't, not to duplicate etc. Please make your submissions early even if they are not feature-complete so that we can give early review on marker stylings like colors, patterns, line widths etc.

The element with EPR regions (such as `Swissmon` in our example) may not necessarily placed at point 0,0. It may also not be oriented in standard way, it might be rotated or mirrored. So the coordinate system of `LayoutView` object may not align with that of the `Swissmon` [Instance](https://www.klayout.de/doc-qt5/code/class_Instance.html). We need to apply a transformation to the EPR shapes we intend to draw with `Marker`s, which we didn't need to do in current implementation. We recommend to retrieve this transformation like so:

```
for inst_path in lay.LayoutView.current().each_object_selected():
    trans = inst_path.inst().dcplx_trans
```
This code will look up which instance was "selected" in KLayout GUI. We assume that an instance must be selected if `Object Properties` window is opened for this instance. Then we get [DCplxTrans](https://www.klayout.de/doc-qt5/code/class_DCplxTrans.html) object via `Instance`. Part of your task is to handle cases where no selection is present or multiple selections are present - to limit issue scope your implementation can refuse to draw anything unless exactly one selection is present - however it shouldn't crash or cause errors in other scenarios.

### Acceptance criteria

- `_epr_cross_section_cut` boolean parameter toggles drawing correction cuts for element when available, with text markers at endpoints of cut
- `_epr_part_red_xxx` boolean parameter toggles drawing polygons for each available partition region of the element, with a text marker in the middle of the partition region.
- Above shapes follow element when it is moved, rotated, mirrored etc
- No marker lingering, duplication bugs

l997RaghavAnand:
fix #125

### DIFF
--- a/klayout_package/newtest.py
+++ b/klayout_package/newtest.py
@@ -1,0 +1,1 @@
+testing


qpavsmi:
Looking forward to your submission. My advice is to take the time to give a meaningful name to the commits that you make, this also helps me to keep track which changes you have made over time.

PR #133, opened 2026-06-03
Add independent coupler ground gaps and improved Swissmon EPR partitioning/correction cuts

AbdullahKazi500:
# Add independent coupler ground gaps and improved Swissmon EPR partitioning/correction cuts

## Summary

This PR improves Swissmon geometry configurability and updates the associated EPR extraction logic accordingly.

The changes include:

- independent coupler gap-to-ground parameters (`cpl_b`)
- improved EPR partitioning of the central cross
- symmetric correction cuts for all cross arms
- dynamic coupler partition region sizing based on waveguide length
- removal of the unintended dependency between coupler geometry and global CPW gap `b`

---

# Changes

### Requirement 3 — `cpl_b` parameter

Added:

```python
cpl_b = Param(pdt.TypeList, "Coupler gap to ground (um, WNE)", [4, 4, 4])
```

fixes #116


AbdullahKazi500:
@qpavsmi Hi pavel any updates on this PR I did an intial attempt on this

qpavsmi:
Changes were made to the wrong file, it should be in https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/epr/swissmon.py

Please also be mindful that the [code changes](https://github.com/iqm-finland/KQCircuits/pull/133/changes) are clean - avoid complete rewrites of the code

AbdullahKazi500:
> Changes were made to the wrong file, it should be in https://github.com/iqm-finland/KQCircuits/blob/main/klayout_package/python/kqcircuits/simulations/epr/swissmon.py
> 
> Please also be mindful that the [code changes](https://github.com/iqm-finland/KQCircuits/pull/133/changes) are clean - avoid complete rewrites of the code

sure BTW What part of the original code is  not to be touched in that case

qpavsmi:
> sure BTW What part of the original code is not to be touched in that case

Just generally when all code is deleted and written from scratch it is more difficult to review the code. It is prefered if individual lines or small blocks are incrementally added/updated. Nothing specific needs to be preserved, except the function headers `partition_regions`, `correction_cuts` etc.

AbdullahKazi500:
> > sure BTW What part of the original code is not to be touched in that case
> 
> Just generally when all code is deleted and written from scratch it is more difficult to review the code. It is prefered if individual lines or small blocks are incrementally added/updated. Nothing specific needs to be preserved, except the function headers `partition_regions`, `correction_cuts` etc.

Thanks I will keep that in mind

AbdullahKazi500:
> > sure BTW What part of the original code is not to be touched in that case
> 
> Just generally when all code is deleted and written from scratch it is more difficult to review the code. It is prefered if individual lines or small blocks are incrementally added/updated. Nothing specific needs to be preserved, except the function headers `partition_regions`, `correction_cuts` etc.

@qpavsmi One thing I want to double check visually in KLayout before finalizing: the ``crosstop`` and ``crossright`` cut geometry  specifically whether epr_cplr1_max and epr_cplr2_max are the correct reference corners for those arms (they should be, given couplers 1=North, 2=East by the rotation logic in ``_produce_coupler``

AbdullahKazi500:
I’ve pushed the requested updates for the Swissmon EPR partition region and correction cut handling. Apologies for the confusion earlier  I got a bit mixed up with the file naming/path and directory structure while tracking the correct file for modifications. if there is any mistake you found in the PR Then please let me know with respect to the path 

# Changes made
1. Split the original cross partition region into 4 arm-specific regions

Replaced the single:

cross

with:

- crossleft
- crosstop
- crossright
- crossbottom

Each arm now creates its own polygon region using the corresponding epr_cross_* points together with the base refpoint.

2. Updated cross MER correction cuts

Replaced the single:

crossmer

with:

- crossleftmer
- crosstopmer
- crossrightmer
- crossbottommer

The west arm (crossleftmer) preserves the original geometry, while the other three are rotated variants using the respective arm reference points.

3. Updated coupler waveguide offsets

Changed _get_coupler_wg_offset() to use:

waveguide_length + 25

instead of the previously hardcoded 150 offsets.

This now adapts dynamically when waveguide_length exists on the simulation object.

4. Updated coupler correction cuts to use per-coupler gap values

Removed:

half_gap = simulation.b / 2

and replaced it with:

half_gap = float(simulation.cpl_b[i]) / 2

inside each coupler block individually (0, 1, 2).

5. Added cpl_b to extracted parameters

Updated:

extract_swissmon_from(...)

to include:

"cpl_b",

while keeping "b" intact since it may still be used elsewhere in the base qubit implementation.

AbdullahKazi500:
also For the new arm-specific MER cuts (crosstopmer, crossrightmer, crossbottommer), should the current naming convention be considered final, or would you prefer a different naming style?
For the dynamic waveguide offset:
waveguide_length + 25

is the additional 25 expected to remain fixed, or should this eventually become parameterized as well?

Would you also want separate handling for varying gap_width values across different cross arms in the future? Currently the existing warning behavior remains unchanged.

qpavsmi:
> also For the new arm-specific MER cuts (crosstopmer, crossrightmer, crossbottommer), should the current naming convention be considered final, or would you prefer a different naming style? For the dynamic waveguide offset: waveguide_length + 25
> 
> is the additional 25 expected to remain fixed, or should this eventually become parameterized as well?
> 
> Would you also want separate handling for varying gap_width values across different cross arms in the future? Currently the existing warning behavior remains unchanged.

Thanks for the submission. On a glance it looks like you have everything needed to fulfill the task. I will address your questions during the review process. I think the 25 can be left unparametrized, but I will know for sure during the review.

I'm a bit backed up by work tasks and other UH submissions so I can't promise that I will review today, hopefully you can wait until Monday to receive feedback.

AbdullahKazi500:
> > also For the new arm-specific MER cuts (crosstopmer, crossrightmer, crossbottommer), should the current naming convention be considered final, or would you prefer a different naming style? For the dynamic waveguide offset: waveguide_length + 25
> > is the additional 25 expected to remain fixed, or should this eventually become parameterized as well?
> > Would you also want separate handling for varying gap_width values across different cross arms in the future? Currently the existing warning behavior remains unchanged.
> 
> Thanks for the submission. On a glance it looks like you have everything needed to fulfill the task. I will address your questions during the review process. I think the 25 can be left unparametrized, but I will know for sure during the review.
> 
> I'm a bit backed up by work tasks and other UH submissions so I can't promise that I will review today, hopefully you can wait until Monday to receive feedback.

Thanks no issues I also had a doctor visit today an accident few months happened with me and I got staph infection so I am recovering from it these days

AbdullahKazi500:
> > also For the new arm-specific MER cuts (crosstopmer, crossrightmer, crossbottommer), should the current naming convention be considered final, or would you prefer a different naming style? For the dynamic waveguide offset: waveguide_length + 25
> > is the additional 25 expected to remain fixed, or should this eventually become parameterized as well?
> > Would you also want separate handling for varying gap_width values across different cross arms in the future? Currently the existing warning behavior remains unchanged.
> 
> Thanks for the submission. On a glance it looks like you have everything needed to fulfill the task. I will address your questions during the review process. I think the 25 can be left unparametrized, but I will know for sure during the review.
> 
> I'm a bit backed up by work tasks and other UH submissions so I can't promise that I will review today, hopefully you can wait until Monday to receive feedback.

Hi @qpavsmi any updates on this ?

klayout_package/python/kqcircuits/qubits/swissmon.py
```diff
@@ -43,6 +43,7 @@ class Swissmon(Qubit):
     cpl_width = Param(pdt.TypeList, "Coupler width (um, WNE)", [24, 24, 24])
     cpl_length = Param(pdt.TypeList, "Coupler lengths (um, WNE)", [120, 120, 120])
     cpl_gap = Param(pdt.TypeList, "Coupler gap (um, WNE)", [102, 102, 102])
+    cpl_b = Param(pdt.TypeList, "Coupler gap to ground (um, WNE)", [4, 4, 4])
```
qpavsmi:
```suggestion
    cpl_b = Param(pdt.TypeList, "Coupler gap to ground (um, WNE)", [6, 6, 6])
```
`b` parameter default is 6, let's have it 6 here as well

qpavsmi:
There are issues with following correction cut definitions on Swissmon qubit with default PCell parameters:

- `crosstopmer` starts from internal cross metal island, crosses over to grounded metal, then crosses over to north coupler region. The cut should not extend all the way to the coupler shape.
- `crossrightmer` is a short cut that starts and ends inside the cross metal island. The idea of correction cuts is that it cuts across a gap region to characterize a metal-gap profile of a particular region.
- `crossbottommer` cut starts ok, but it ends at cross metal island. It should extend to the other side of the cross metal island.

Please compare to how the `crossleftmer` cut defined, which is what we want. Starts from grounded metal outside of cross metal, crosses through cross metal, then ends on the other side outside of cross metal at grounded metal. Please see below illustration, where I highlighted in red cuts that need to be fixed and in green a correction cut that is correct.

image

I suggest that you move the correction cut closed to the center, this will decrease the likelihood that a coupler will get in the way of the correction cut like it does for `crosstopmer`. The cuts are allowed to overlap with each other: this is not an issue and you shouldn't worry about it. For example `crosstopmer` cut is allowed to cross through `crossleftmer` cut. Below I illustrated how I envision the cuts could be defined:

image

Make sure that when you try different parameters for Swissmon, the correction cuts don't violate above listed issues. You can see that `crossleftmer` will adjust its location if `cpl_length` is increased so that it will not cut into the coupler shape. Try implementing similar to your three other cuts as well.

AbdullahKazi500:
Hi @qpavsmi thanks for the review I agree with you my question is 
- When moving the cuts closer to the center, is there a preferred “safe margin” from the internal cross island edges, or is an approximate symmetric placement acceptable as long as the cut crosses:
grounded metal → gap/metal region → grounded metal?
- For crossrightmer, should the correction cut be centered similarly to crossleftmer, i.e. midway between the inner shoulder and outer arm edge, rather than near the arm tip?
- For crossbottommer, do you want the cut to extend equally beyond both sides of the cross island, or should one side adapt dynamically like crossleftmer depending on nearby structures?

qpavsmi:
> Hi @qpavsmi thanks for the review I agree with you my question is
> 
> * When moving the cuts closer to the center, is there a preferred “safe margin” from the internal cross island edges, or is an approximate symmetric placement acceptable as long as the cut crosses:
>   grounded metal → gap/metal region → grounded metal?
> * For crossrightmer, should the correction cut be centered similarly to crossleftmer, i.e. midway between the inner shoulder and outer arm edge, rather than near the arm tip?
> * For crossbottommer, do you want the cut to extend equally beyond both sides of the cross island, or should one side adapt dynamically like crossleftmer depending on nearby structures?

- Looking at original code for `crossleftmer` the `half_cut_length` is defined as 30 µm + the width of the cross arm. You can measure the cut in KLayout that it extends 30 µm on both sides of the cross arm. I think this is acceptable margin. It can be shorter, but there should be some portion of the ground metal on both sides of the cut.
- Please make `crossrightmer` resemble `crossleftmer` as much as you can, as long as it is located on the opposite arm. That is better to be closer to center rather than on the tip of the arm.
- Make it adapt if you can. The `crossleftmer` should be a standard for all `cross*mer` cuts when it comes to placement and adaptability to Swissmon parameters.

AbdullahKazi500:
> > Hi @qpavsmi thanks for the review I agree with you my question is
> > 
> > * When moving the cuts closer to the center, is there a preferred “safe margin” from the internal cross island edges, or is an approximate symmetric placement acceptable as long as the cut crosses:
> >   grounded metal → gap/metal region → grounded metal?
> > * For crossrightmer, should the correction cut be centered similarly to crossleftmer, i.e. midway between the inner shoulder and outer arm edge, rather than near the arm tip?
> > * For crossbottommer, do you want the cut to extend equally beyond both sides of the cross island, or should one side adapt dynamically like crossleftmer depending on nearby structures?
> 
> * Looking at original code for `crossleftmer` the `half_cut_length` is defined as 30 µm + the width of the cross arm. You can measure the cut in KLayout that it extends 30 µm on both sides of the cross arm. I think this is acceptable margin. It can be shorter, but there should be some portion of the ground metal on both sides of the cut.
> * Please make `crossrightmer` resemble `crossleftmer` as much as you can, as long as it is located on the opposite arm. That is better to be closer to center rather than on the tip of the arm.
> * Make it adapt if you can. The `crossleftmer` should be a standard for all `cross*mer` cuts when it comes to placement and adaptability to Swissmon parameters.

- thanks pavel I am working on this I am also confused with the overlapping between correction cuts  is this always acceptable, even for extreme parameter combinations (large coupler lengths / asymmetric arm widths), as long as they do not intersect coupler metal regions?
- 
- Should I also test the cuts against non-default Swissmon parameters (for example asymmetric arm lengths/gaps/coupler lengths), or is validating against a few representative parameter combinations sufficient?

qpavsmi:
> * thanks pavel I am working on this I am also confused with the overlapping between correction cuts  is this always acceptable, even for extreme parameter combinations (large coupler lengths / asymmetric arm widths), as long as they do not intersect coupler metal regions?
> * Should I also test the cuts against non-default Swissmon parameters (for example asymmetric arm lengths/gaps/coupler lengths), or is validating against a few representative parameter combinations sufficient?

- You should only worry that a correction cut crosses the metal and gap regions as described previously. You should not worry about multiple correction cuts crossing over each other - that does not affect simulations in any way. If some parameter value causes two correction cuts to overlap - no problem. If some parameter value causes some correction cut to cross to the couplers, or to not cross over from ground metal, to cross arm, back to ground metal - there is a problem.
- Check with reasonable parameter variations that the cuts and partition regions are ok. But there are parameter combinations that break Swissmon shape and we shouldn't worry about correction cuts and partition regions in such cases. For example if some value in `cpl_length` is so big that the coupler starts overlapping with the cross, or comes so close to the cross that a correction cut cannot fit, then disregard such parameter combination as it is a faulty design anyway. A +- 20µm variation of a parameter is probably good to support.

klayout_package/python/kqcircuits/simulations/epr/swissmon.py
```diff
@@ -29,30 +34,55 @@
...
+
+    for arm_name, indices in [
+        ("crossleft", [6, 7, 8, 9]),
+        ("crosstop", [9, 10, 11, 0]),
+        ("crossright", [0, 1, 2, 3]),
+        ("crossbottom", [3, 4, 5, 6]),
+    ]:
+        arm_poly = pya.DPolygon(
+            [simulation.refpoints[f"epr_cross_{i:02d}"] for i in indices] + [base]
+        ).sized(sized)
```
qpavsmi: You defined the corner points of each cross polygon correctly, and you also sized the polygon correctly. But it's better not to size the polygon near the `base` refpoint. This causes the polygons to overlap with each other. See illustration below to see what I mean:

Image

My suggestion is:
- Construct `pya.DPolygon` using `epr_cross_*` refpoints **but without** `base`
- Use `.sized` function on polygon
- Use `point_hull` function to get the points from sized polygon
- Construct new `pya.DPolygon` using extracted points from previous step, then add `base` refpoint such that it wasn't subject to `.sized` function, use that polygon.

qpavsmi (changes requested):
In addition to the discussion about correction cuts across swissmon arms, here are the changes I request for this submission

qpavsmi:
> also For the new arm-specific MER cuts (crosstopmer, crossrightmer, crossbottommer), should the current naming convention be considered final, or would you prefer a different naming style? For the dynamic waveguide offset: waveguide_length + 25
> 
> is the additional 25 expected to remain fixed, or should this eventually become parameterized as well?
> 
> Would you also want separate handling for varying gap_width values across different cross arms in the future? Currently the existing warning behavior remains unchanged.

The naming convention is OK. Static 25µm is OK at this time. I will test at a later time if varying `gap_width` breaks something. Generally my policy is if it breaks something in `main` branch, then it is ok to break in your branch, unless in review I am able to suggest some easy fix.

Let me know if I answered all your questions at this time of development.

AbdullahKazi500:
> > also For the new arm-specific MER cuts (crosstopmer, crossrightmer, crossbottommer), should the current naming convention be considered final, or would you prefer a different naming style? For the dynamic waveguide offset: waveguide_length + 25
> > is the additional 25 expected to remain fixed, or should this eventually become parameterized as well?
> > Would you also want separate handling for varying gap_width values across different cross arms in the future? Currently the existing warning behavior remains unchanged.
> 
> The naming convention is OK. Static 25µm is OK at this time. I will test at a later time if varying `gap_width` breaks something. Generally my policy is if it breaks something in `main` branch, then it is ok to break in your branch, unless in review I am able to suggest some easy fix.
> 
> Let me know if I answered all your questions at this time of development.

Hi @qpavsmi I am making the changes feel free to let me know

AbdullahKazi500:
also I noticed in one of the review comments/examples that ``cpl_b`` was shown changing from:
``cpl_b = Param(pdt.TypeList, "Coupler gap to ground (um, WNE)", [4, 4, 4])``
to
``cpl_b = Param(pdt.TypeList, "Coupler gap to ground (um, WNE)", [6, 6, 6])``

I do not see this change mentioned in the issue requirements themselves, so I wanted to clarify whether this was only an example used while testing correction-cut robustness or whether support for a larger cpl_b value is an expected part of the fix. My current understanding is that changing cpl_b modifies the coupler gap-to-ground spacing and therefore affects the geometry around the couplers, but should not be necessary for the correction-cut implementation itself. Could you confirm whether the intended solution should work with the existing default value [4, 4, 4], or whether there is a specific reason to prefer [6, 6, 6] for this issue?

qpavsmi:
> also I noticed in one of the review comments/examples that `cpl_b` was shown changing from: `cpl_b = Param(pdt.TypeList, "Coupler gap to ground (um, WNE)", [4, 4, 4])` to `cpl_b = Param(pdt.TypeList, "Coupler gap to ground (um, WNE)", [6, 6, 6])`
> 
> I do not see this change mentioned in the issue requirements themselves, so I wanted to clarify whether this was only an example used while testing correction-cut robustness or whether support for a larger cpl_b value is an expected part of the fix. My current understanding is that changing cpl_b modifies the coupler gap-to-ground spacing and therefore affects the geometry around the couplers, but should not be necessary for the correction-cut implementation itself. Could you confirm whether the intended solution should work with the existing default value [4, 4, 4], or whether there is a specific reason to prefer [6, 6, 6] for this issue?

If you set the value to `[4, 4, 4]` then you change the default shape of Swissmon. Notice that default value of `b` is 6. That means for the default shape of Swissmon to stay the same as in `main` branch the default value should be `[6, 6, 6]`

AbdullahKazi500:
> > also I noticed in one of the review comments/examples that `cpl_b` was shown changing from: `cpl_b = Param(pdt.TypeList, "Coupler gap to ground (um, WNE)", [4, 4, 4])` to `cpl_b = Param(pdt.TypeList, "Coupler gap to ground (um, WNE)", [6, 6, 6])`
> > I do not see this change mentioned in the issue requirements themselves, so I wanted to clarify whether this was only an example used while testing correction-cut robustness or whether support for a larger cpl_b value is an expected part of the fix. My current understanding is that changing cpl_b modifies the coupler gap-to-ground spacing and therefore affects the geometry around the couplers, but should not be necessary for the correction-cut implementation itself. Could you confirm whether the intended solution should work with the existing default value [4, 4, 4], or whether there is a specific reason to prefer [6, 6, 6] for this issue?
> 
> If you set the value to `[4, 4, 4]` then you change the default shape of Swissmon. Notice that default value of `b` is 6. That means for the default shape of Swissmon to stay the same as in `main` branch the default value should be `[6, 6, 6]`

Okay sure can you review the changes ? I made recently

qpavsmi:
> Okay sure can you review the changes ? I made recently

You messed up your files now, the `klayout_package/python/kqcircuits/qubits/swissmon.py` file has code intended for `klayout_package/python/kqcircuits/simulations/epr/swissmon.py`

Please be more careful when using the git tool before pushing new changes.

AbdullahKazi500:
> > Okay sure can you review the changes ? I made recently
> 
> You messed up your files now, the `klayout_package/python/kqcircuits/qubits/swissmon.py` file has code intended for `klayout_package/python/kqcircuits/simulations/epr/swissmon.py`
> 
> Please be more careful when using the git tool before pushing new changes.

yes apologies I wanted to tell you that I did to be very honest both the names are very similar To be honest made a mistake but rest apart does everything looks good

qpavsmi:
> yes apologies I wanted to tell you that I did to be very honest both the names are very similar To be honest made a mistake but rest apart does everything looks good

Please fix. I hope you can understand that it is difficult for me to review code if it is in a messy state.

AbdullahKazi500:
> > yes apologies I wanted to tell you that I did to be very honest both the names are very similar To be honest made a mistake but rest apart does everything looks good
> 
> Please fix. I hope you can understand that it is difficult for me to review code if it is in a messy state.

sure will fix it now

AbdullahKazi500:
> > yes apologies I wanted to tell you that I did to be very honest both the names are very similar To be honest made a mistake but rest apart does everything looks good
> 
> Please fix. I hope you can understand that it is difficult for me to review code if it is in a messy state.

Just fixed the files actually I got confused within the comments I hope it looks better if there is any fix in ``klayout_package/python/kqcircuits/qubits/swissmon.py`` then let me know thanks

klayout_package/python/kqcircuits/simulations/epr/swissmon.py
```diff
@@ -1,5 +1,5 @@
 # This code is part of KQCircuits
-# Copyright (C) 2024 IQM Finland Oy
+# Copyright (C) 2021 IQM Finland Oy
```
qpavsmi: No reason to change copyright year

AbdullahKazi500:
also one questions 
`` cpl_b = Param(pdt.TypeList, "Coupler gap to ground (um, WNE)", [6, 6, 6])``
this was the only change for this file if I am right 

[klayout_package/python/kqcircuits/qubits/swissmon.py]

qpavsmi:
The cuts can be moved closer to center still. Please see how close `crossleftmer` is to center and make other `cross*mer` cuts as close to center. Remember that it is ok if cuts cross over each other.

Currently `crossrightmer` and `crosstopmer` cut into couplers on swissmon with default parameters. We want to avoid this.

image

image

qpavsmi:
image

`crossleftbulk`, `crossrightbulk`, `crossbottombulk` regions are broken. Check that the order of points in polygon are correct.

klayout_package/python/kqcircuits/simulations/epr/swissmon.py
```diff
@@ -29,30 +34,61 @@
...
+    ]:
+        # Build polygon from epr_cross_* refpoints only (without base), size it,
+        # then reconstruct adding base unsized so the four arm polygons meet at center
+        # without overlapping each other.
+        arm_points_poly = pya.DPolygon(
+            [simulation.refpoints[f"epr_cross_{i:02d}"] for i in indices]
+        )
+        sized_poly = arm_points_poly.sized(sized)
+        hull_points = [sized_poly.point_hull(i) for i in range(sized_poly.num_points_hull())]
+        arm_poly = pya.DPolygon(hull_points + [base])
```
qpavsmi: I realize now that this approach (suggested by me) probably doesn't work. I think instead, you should take the two point that are away from cross center (`epr_cross_01`, `epr_cross_02`, `epr_cross_04`, `epr_cross_05`, `epr_cross_01`, `epr_cross_07`, `epr_cross_08`, `epr_cross_10`, `epr_cross_11`), and move them by `sized` value on X and Y axis away from origin. You can use `+` operator to do this. Then the two points closer to center (`epr_cross_00`, `epr_cross_03`, `epr_cross_06`, `epr_cross_09`) should stay as they are, and eventually connected to `base` refpoint.

qpavsmi (changes requested):
I think the polygon resizing approach should be revisited

klayout_package/python/kqcircuits/simulations/epr/swissmon.py
qpavsmi: This issue probably is related to https://github.com/iqm-finland/KQCircuits/pull/133#issuecomment-4669744130

AbdullahKazi500: > This issue probably is related to [#133 (comment)](https://github.com/iqm-finland/KQCircuits/pull/133#issuecomment-4669744130)

issue regarding what ?

AbdullahKazi500:
Hi @qpavsmi I am kind of need some guidance on the following questions after your recent comment


- Regarding the correction cuts, should I move crosstopmer, crossrightmer, and crossbottommer much closer to the center of the cross, using crossleftmer as the reference for the intended placement?
- For avoiding coupler intersections, should the logic be similar to crossleftmer, where the cut position adapts based on the coupler extent (epr_cplr*_max) when a coupler is present?
- For crosstopmer, should the cut location be determined primarily from the inner cross geometry and only secondarily constrained by the coupler position, so that it remains close to the center even when the coupler length changes?
- Similarly for crossrightmer, should I mirror the same strategy used by crossleftmer, placing the cut between the inner cross edge and the coupler while ensuring it never enters the coupler region?
- Regarding the partition-region comments, I noticed your follow-up note that the sized_poly/point_hull() approach may not be correct. Am I understanding correctly that I should abandon that approach entirely and instead manually offset only the outer refpoints (01, 02, 04, 05, 07, 08, 10, 11) away from the origin by the sizing amount?

klayout_package/python/kqcircuits/simulations/epr/swissmon.py
AbdullahKazi500: > This issue probably is related to [#133 (comment)](https://github.com/iqm-finland/KQCircuits/pull/133#issuecomment-4669744130)

okay issue with respect to the image I see

AbdullahKazi500:
apologies for asking too many questions 
- For those outer refpoints, should the offset be applied independently along both X and Y according to the sign of the coordinate (i.e., move positive coordinates further positive and negative coordinates further negative)?
- Should the inner refpoints (00, 03, 06, 09) remain completely unchanged and continue to connect directly to the base refpoint?

klayout_package/python/kqcircuits/simulations/epr/swissmon.py
AbdullahKazi500: yeah sorry my mistake will fix it

qpavsmi:
> Hi @qpavsmi I am kind of need some guidance on the following questions after your recent comment
> 
> * Regarding the correction cuts, should I move crosstopmer, crossrightmer, and crossbottommer much closer to the center of the cross, using crossleftmer as the reference for the intended placement?

Yes
> * For avoiding coupler intersections, should the logic be similar to crossleftmer, where the cut position adapts based on the coupler extent (epr_cplr*_max) when a coupler is present?

Yes
> * For crosstopmer, should the cut location be determined primarily from the inner cross geometry and only secondarily constrained by the coupler position, so that it remains close to the center even when the coupler length changes?

A bit difficult to understand the question. As long as it's as close to center as `crossleftmer` is, it should be fine. If you're worried about some particular parameter combo, share a screenshot with me and I can decide from there.

> * Similarly for crossrightmer, should I mirror the same strategy used by crossleftmer, placing the cut between the inner cross edge and the coupler while ensuring it never enters the coupler region?

Yes
> * Regarding the partition-region comments, I noticed your follow-up note that the sized_poly/point_hull() approach may not be correct. Am I understanding correctly that I should abandon that approach entirely and instead manually offset only the outer refpoints (01, 02, 04, 05, 07, 08, 10, 11) away from the origin by the sizing amount?

If you could manually offset the points further from center. I realise now that sizing the entire polygon would also offset the points closer to center. But I also now realise that the points close to center should be adjusted, but in opposite direction to where they are now.

Please use your judgement to implement this so that the outcome is as follows:

image

> * For those outer refpoints, should the offset be applied independently along both X and Y according to the sign of the coordinate (i.e., move positive coordinates further positive and negative coordinates further negative)?

Do what you can so the outcome is as above picture
> * Should the inner refpoints (00, 03, 06, 09) remain completely unchanged and continue to connect directly to the base refpoint?

Probably shouldn't be unchanged (my previous suggestion was wrong). See above picture for what the outcome should be.

AbdullahKazi500:
> > Hi @qpavsmi I am kind of need some guidance on the following questions after your recent comment
> > 
> > * Regarding the correction cuts, should I move crosstopmer, crossrightmer, and crossbottommer much closer to the center of the cross, using crossleftmer as the reference for the intended placement?
> 
> Yes
> 
> > * For avoiding coupler intersections, should the logic be similar to crossleftmer, where the cut position adapts based on the coupler extent (epr_cplr*_max) when a coupler is present?
> 
> Yes
> 
> > * For crosstopmer, should the cut location be determined primarily from the inner cross geometry and only secondarily constrained by the coupler position, so that it remains close to the center even when the coupler length changes?
> 
> A bit difficult to understand the question. As long as it's as close to center as `crossleftmer` is, it should be fine. If you're worried about some particular parameter combo, share a screenshot with me and I can decide from there.
> 
> > * Similarly for crossrightmer, should I mirror the same strategy used by crossleftmer, placing the cut between the inner cross edge and the coupler while ensuring it never enters the coupler region?
> 
> Yes
> 
> > * Regarding the partition-region comments, I noticed your follow-up note that the sized_poly/point_hull() approach may not be correct. Am I understanding correctly that I should abandon that approach entirely and instead manually offset only the outer refpoints (01, 02, 04, 05, 07, 08, 10, 11) away from the origin by the sizing amount?
> 
> If you could manually offset the points further from center. I realise now that sizing the entire polygon would also offset the points closer to center. But I also now realise that the points close to center should be adjusted, but in opposite direction to where they are now.
> 
> Please use your judgement to implement this so that the outcome is as follows:
> 
> image
> > * For those outer refpoints, should the offset be applied independently along both X and Y according to the sign of the coordinate (i.e., move positive coordinates further positive and negative coordinates further negative)?
> 
> Do what you can so the outcome is as above picture
> 
> > * Should the inner refpoints (00, 03, 06, 09) remain completely unchanged and continue to connect directly to the base refpoint?
> 
> Probably shouldn't be unchanged (my previous suggestion was wrong). See above picture for what the outcome should be.

Ok sure I will try my best and get back to you I hope we can fix this issue soon thanks for the review

AbdullahKazi500:
Hi @qpavsmi made a couple of modifications do let me know

qpavsmi:
The cuts are too close now. They don't cut across any gaps in the swissmon qubit, they only stay on metal or tangentially touch the gaps.

image

Please pay closer attention to my previous suggestions, most importantly this illustration I made previously

image

The `crossleftmer` cut location was fine in previous commits. There was no reason to move it.

qpavsmi:
image

The inner points of partition regions should be moved outwards a bit. The idea is that regions have straight lines along X and Y axises, not crooked lines.

AbdullahKazi500:
Ok sure @qpavsmi no issues

AbdullahKazi500:
made some changes @qpavsmi

qpavsmi:
> made some changes @qpavsmi

All I see in recent commit (https://github.com/iqm-finland/KQCircuits/pull/133/changes/459f37f283878fac9c94178fcbb011bc63a52944) are changes in code comments, not in code itself. I still get same bugs that I reported previously:
https://github.com/iqm-finland/KQCircuits/pull/133#issuecomment-4677837918
https://github.com/iqm-finland/KQCircuits/pull/133#issuecomment-4677903236

AbdullahKazi500:
> > made some changes @qpavsmi
> 
> All I see in recent commit ([459f37f](https://github.com/iqm-finland/KQCircuits/commit/459f37f283878fac9c94178fcbb011bc63a52944)) are changes in code comments, not in code itself. I still get same bugs that I reported previously: [#133 (comment)](https://github.com/iqm-finland/KQCircuits/pull/133#issuecomment-4677837918) [#133 (comment)](https://github.com/iqm-finland/KQCircuits/pull/133#issuecomment-4677903236)

weird I made the changes in the code and added an updated PR Oh I see the issue was with the internet sorry the PR is on the way no issues

AbdullahKazi500:
> > made some changes @qpavsmi
> 
> All I see in recent commit ([459f37f](https://github.com/iqm-finland/KQCircuits/commit/459f37f283878fac9c94178fcbb011bc63a52944)) are changes in code comments, not in code itself. I still get same bugs that I reported previously: [#133 (comment)](https://github.com/iqm-finland/KQCircuits/pull/133#issuecomment-4677837918) [#133 (comment)](https://github.com/iqm-finland/KQCircuits/pull/133#issuecomment-4677903236)

Hi Pavel as I mentioned I made the PR Earlier but was disrupted due to the internet so added an updated PR now with the fixes

qpavsmi:
I'm still seeing the problem with partition regions: https://github.com/iqm-finland/KQCircuits/pull/133#issuecomment-4677903236

The correction cuts seem to have changed, reviewing them

AbdullahKazi500:
> I'm still seeing the problem with partition regions: [#133 (comment)](https://github.com/iqm-finland/KQCircuits/pull/133#issuecomment-4677903236)
> 
> The correction cuts seem to have changed, reviewing them

yes please this is all the fix 

partition_regions unchanged, _offset_point_toward on inner corners and _offset_point_away on outer points is already correct per the maintainer's guidance.
correction_cuts — three cuts were wrong:

crosstopmer: was using epr_cplr1_max (outer face of coupler, far from center) as the coupler anchor — changed to epr_cplr1_min (the center-facing inner face), mirroring exactly what crossleftmer does with epr_cplr0_max
crossrightmer: same fix changed from epr_cplr2_max to epr_cplr2_min so the midpoint sits between the inner gap edge and the coupler's center-facing wall, not its far wall
crossbottommer: was using epr_cross_04.y in an average with cross_corner_bot_l.y — kept this correctly as the midpoint between inner gap edge and arm tip, which is the right analog of crossleftmer's formula for the no-coupler case

qpavsmi:
> yes please this is all the fix
> 
> partition_regions unchanged, _offset_point_toward on inner corners and _offset_point_away on outer points is already correct per the maintainer's guidance. correction_cuts — three cuts were wrong:

Partition regions are not correct. See the picture again that I posted yesterday

image

There should be no skewed lines, they should be aligned on X and Y-axis

AbdullahKazi500:
> > yes please this is all the fix
> > partition_regions unchanged, _offset_point_toward on inner corners and _offset_point_away on outer points is already correct per the maintainer's guidance. correction_cuts — three cuts were wrong:
> 
> Partition regions are not correct. See the picture again that I posted yesterday
> 
> image
> There should be no skewed lines, they should be aligned on X and Y-axis

Okay I tried a lot of things do you have any suggestions or hints for this

klayout_package/python/kqcircuits/simulations/epr/swissmon.py
```diff
@@ -16,42 +16,99 @@
...
+        ("crossright",  [0, 1, 2, 3]),
+        ("crossbottom", [3, 4, 5, 6]),
+    ]:
+        raw = [simulation.refpoints[f"epr_cross_{i:02d}"] for i in indices]
+
+        pts = [
+            _offset_point_toward(raw[0], sized),  # inner corner A → toward origin
+            _offset_point_away(raw[1], sized),     # outer point   → away from origin
+            _offset_point_away(raw[2], sized),     # outer point   → away from origin
+            _offset_point_toward(raw[3], sized),   # inner corner B → toward origin
```
qpavsmi:
```suggestion
            _offset_point_away(raw[0], sized),
            _offset_point_away(raw[1], sized),
            _offset_point_away(raw[2], sized),
            _offset_point_away(raw[3], sized),
```
This is the change needed to fix the partition regions, `_offset_point_toward` can be deleted

AbdullahKazi500:
## I applied the suggestion you made thank you 

BTW from my point of view and understanding The old code built each arm as a pya.DPolygon(pts + [base])  a pentagon where the 5th vertex was the origin (0, 0). This connected the two inner gap corners to the origin with diagonal edges, which is exactly the wedge shapes visible in the image. or maybe I am wrong I tried stuff

qpavsmi:
> ## I applied the suggestion you made thank you
> BTW from my point of view and understanding The old code built each arm as a pya.DPolygon(pts + [base]) a pentagon where the 5th vertex was the origin (0, 0). This connected the two inner gap corners to the origin with diagonal edges, which is exactly the wedge shapes visible in the image. or maybe I am wrong I tried stuff

The earlier implementation caused the partition regions to overlap, now they are not overlapping

AbdullahKazi500:
> > ## I applied the suggestion you made thank you
> > BTW from my point of view and understanding The old code built each arm as a pya.DPolygon(pts + [base]) a pentagon where the 5th vertex was the origin (0, 0). This connected the two inner gap corners to the origin with diagonal edges, which is exactly the wedge shapes visible in the image. or maybe I am wrong I tried stuff
> 
> The earlier implementation caused the partition regions to overlap, now they are not overlapping

okay No issue let me know if anything more needs to be done

klayout_package/python/kqcircuits/simulations/epr/swissmon.py
```diff
@@ -16,42 +16,99 @@
...
 
 
 # Partition region and correction cuts definitions for Swissmon qubit
 
 
+def _offset_point_away(p: pya.DPoint, sized: float) -> pya.DPoint:
+    """Offset a point *away* from the origin by `sized` on each axis."""
+    dx = math.copysign(sized, p.x) if p.x != 0 else 0.0
+    dy = math.copysign(sized, p.y) if p.y != 0 else 0.0
+    return pya.DPoint(p.x + dx, p.y + dy)
```
qpavsmi:
```suggestion
def _offset_point_away(p: pya.DPoint, sized: float, origin: pya.DPoint) -> pya.DPoint:
    """Offset a point *away* from the origin by `sized` on each axis."""
    dx = math.copysign(sized, p.x - origin.x) if p.x != 0 else 0.0
    dy = math.copysign(sized, p.y - origin.y) if p.y != 0 else 0.0
    return pya.DPoint(p.x + dx, p.y + dy)
```
The `p.x` and `p.y` coordinates need to be normalized. Here, the value of `origin` argument should be `base = simulation.refpoints["base"]`.

This makes a difference because if you run `python klayout_package/python/scripts/simulations/swissmon_epr_sim.py`, the partition region will skew incorrectly because swissmon is placed at positive X and Y coordinate, not at 0,0 coordinate.

qpavsmi: Please don't apply the suggestions blindly, but take a moment to integrate the suggestion carefully. The function gets a new argument added to `_offset_point_away`, but in the `_offset_point_away` calls below you don't pass anything for that argument.

AbdullahKazi500: Ok sure let me check once

AbdullahKazi500:
with respect to comment ----https://github.com/iqm-finland/KQCircuits/pull/133#discussion_r3403702509 I would like to make the following changes as suggested by you 
# Requested Changes

## 1. Update `_offset_point_away`

The helper function should now take an `origin` parameter and compute the offset direction relative to that origin rather than the absolute coordinate system.

### Before

```python
def _offset_point_away(p: pya.DPoint, sized: float) -> pya.DPoint:
    dx = math.copysign(sized, p.x) if p.x != 0 else 0.0
    dy = math.copysign(sized, p.y) if p.y != 0 else 0.0
    return pya.DPoint(p.x + dx, p.y + dy)
```

### After

```python
def _offset_point_away(p: pya.DPoint, sized: float, origin: pya.DPoint) -> pya.DPoint:
    """Offset a point away from the origin by `sized` on each axis."""
    dx = math.copysign(sized, p.x - origin.x) if p.x != 0 else 0.0
    dy = math.copysign(sized, p.y - origin.y) if p.y != 0 else 0.0
    return pya.DPoint(p.x + dx, p.y + dy)
```

### Reason

The offset direction should be determined relative to the qubit's actual placement position (`base`) rather than the global coordinate origin `(0, 0)`. This ensures partition region expansion behaves correctly when the element is not centered at the layout origin.

---

## 2. Update All `_offset_point_away` Call Sites

Since `_offset_point_away` now requires an `origin` parameter, all usages in `partition_regions()` should pass the already available `base` refpoint.

### Before

```python
pts = [
    _offset_point_away(raw[0], sized),
    _offset_point_away(raw[1], sized),
    _offset_point_away(raw[2], sized),
    _offset_point_away(raw[3], sized),
]
```

### After

```python
pts = [
    _offset_point_away(raw[0], sized, base),
    _offset_point_away(raw[1], sized, base),
    _offset_point_away(raw[2], sized, base),
    _offset_point_away(raw[3], sized, base),
]
```

### Reason

`base = simulation.refpoints["base"]` is already computed at the beginning of `partition_regions()`, making it the natural reference point for determining the outward direction of the partition region expansion.

---


The change consists of two focused updates:

1. **`_offset_point_away`**

   * Add `origin: pya.DPoint` parameter.
   * Change offset direction calculation from:

     ```python
     math.copysign(sized, p.x)
     ```

     and

     ```python
     math.copysign(sized, p.y)
     ```

     to:

     ```python
     math.copysign(sized, p.x - origin.x)
     ```

     and

     ```python
     math.copysign(sized, p.y - origin.y)
     ```

2. **`partition_regions`**

   * Pass `base` as the third argument to every `_offset_point_away(...)` call.

let me know if this is what you are expecting in an updated PR rather than applying the suggestion ?

qpavsmi:
> with respect to comment ----[#133 (comment)](https://github.com/iqm-finland/KQCircuits/pull/133#discussion_r3403702509)

yes thank you

AbdullahKazi500:
> > with respect to comment ----[#133 (comment)](https://github.com/iqm-finland/KQCircuits/pull/133#discussion_r3403702509)
> 
> yes thank you

updated the PR

qpavsmi:
This submission is feature complete, it fulfills the requirements of the issue. I would still like to have a better look at the code style and code construction after the weekend. It is almost certain that this submission will be accepted, but might need a few revisions when it comes to code style.

AbdullahKazi500:
> This submission is feature complete, it fulfills the requirements of the issue. I would still like to have a better look at the code style and code construction after the weekend. It is almost certain that this submission will be accepted, but might need a few revisions when it comes to code style.

thanks @qpavsmi appreciate thank you for the suggestions , reviews and for bearing with me and my questions apologies for anything happy weekend

AbdullahKazi500:
@qpavsmi you can let me know what final changes I need to make thanks

klayout_package/python/kqcircuits/simulations/epr/swissmon.py
qpavsmi: Either keep this unchanged (2024) or change to 2026

klayout_package/python/kqcircuits/simulations/epr/swissmon.py
```diff
@@ -61,70 +118,156 @@ def _get_coupler_wg_offset(i, s):
...
             "gap_width",
-            "face_ids",  # Accesses list's index 0
+            "face_ids",
             "island_r",
-            "cpl_gap",  # Accesses list's indices 0, 1, 2
-            "cpl_width",  # Accesses list's indices 0, 1, 2
-            "cpl_length",  # Accesses list's indices 0, 1, 2
+            "cpl_gap",
+            "cpl_width",
+            "cpl_length",
```
qpavsmi:
```suggestion
            "cpl_b",  # Accesses list's indices 0, 1, 2
            "gap_width",
            "face_ids",  # Accesses list's index 0
            "island_r",
            "cpl_gap",  # Accesses list's indices 0, 1, 2
            "cpl_width",  # Accesses list's indices 0, 1, 2
            "cpl_length",  # Accesses list's indices 0, 1, 2
```
Better to document which indices are used for array parameters in the code

klayout_package/python/kqcircuits/simulations/epr/swissmon.py
```diff
@@ -16,42 +16,99 @@
...
+    dx = math.copysign(sized, p.x - origin.x) if p.x != 0 else 0.0
+    dy = math.copysign(sized, p.y - origin.y) if p.y != 0 else 0.0
+    return pya.DPoint(p.x + dx, p.y + dy)
+
+
+def _offset_point_toward(p: pya.DPoint, sized: float) -> pya.DPoint:
+    """Offset a point *toward* the origin by `sized` on each axis."""
+    dx = math.copysign(sized, p.x) if p.x != 0 else 0.0
+    dy = math.copysign(sized, p.y) if p.y != 0 else 0.0
+    return pya.DPoint(p.x - dx, p.y - dy)
```
qpavsmi: Remove unused code

klayout_package/python/kqcircuits/simulations/epr/swissmon.py
```diff
@@ -16,42 +16,99 @@
...
 
 
 # Partition region and correction cuts definitions for Swissmon qubit
 
 
+def _offset_point_away(p: pya.DPoint, sized: float, origin: pya.DPoint) -> pya.DPoint:
+    """Offset a point *away* from the origin by `sized` on each axis."""
+    dx = math.copysign(sized, p.x - origin.x) if p.x != 0 else 0.0
+    dy = math.copysign(sized, p.y - origin.y) if p.y != 0 else 0.0
+    return pya.DPoint(p.x + dx, p.y + dy)
```
qpavsmi:
```suggestion
def _offset_point_away(p: pya.DPoint, sized: float, origin: pya.DPoint) -> pya.DPoint:
    """Offset a point *away* from the origin by `sized` on each axis."""
    dx = math.copysign(sized, p.x - origin.x)
    dy = math.copysign(sized, p.y - origin.y)
    return pya.DPoint(p.x + dx, p.y + dy)
```
if statement just complicates the code, we can assume that the refpoints won't be placed on the 0-axis of the swissmon qubit

klayout_package/python/kqcircuits/simulations/epr/swissmon.py
```diff
@@ -16,42 +16,99 @@
...
+    ]:
+        raw = [simulation.refpoints[f"epr_cross_{i:02d}"] for i in indices]
+
+        pts = [
+            _offset_point_away(raw[0], sized, base),
+            _offset_point_away(raw[1], sized, base),
+            _offset_point_away(raw[2], sized, base),
+            _offset_point_away(raw[3], sized, base),
+        ]
+        arm_poly = pya.DPolygon(pts + [base])
```
qpavsmi:
```suggestion
        arm_poly = pya.DPolygon([_offset_point_away(p, sized, base) for p in raw] + [base])
```

klayout_package/python/kqcircuits/simulations/epr/swissmon.py
```diff
@@ -16,42 +16,99 @@
...
+    #
+    # Per-arm index roles [inner_A, outer_1, outer_2, inner_B]:
+    #   crossleft   [6, 7, 8, 9]  — outer: 7, 8   inner: 6, 9
+    #   crosstop    [9,10,11, 0]  — outer:10,11   inner: 9, 0
+    #   crossright  [0, 1, 2, 3]  — outer: 1, 2   inner: 0, 3
+    #   crossbottom [3, 4, 5, 6]  — outer: 4, 5   inner: 3, 6
+    #
+    # Outer points (middle two): offset *away* from origin → expands arm tip outward.
+    # Inner corner points (first and last): offset *toward* origin → pushes inner
+    #   corners inward so arm rectangles overlap at center with straight X/Y edges.
```
qpavsmi:
```suggestion
    # Each arm polygon is a pentagon: 4 offset epr_cross_* points + base (origin).
```
Shorten the comments. Most of this is self-explanatory from code. Also a lot of this is out-of-date

klayout_package/python/kqcircuits/simulations/epr/swissmon.py
```diff
@@ -61,70 +118,156 @@ def _get_coupler_wg_offset(i, s):
...
 
 
 def correction_cuts(simulation: EPRTarget, prefix: str = "") -> dict[str, dict]:
+    # Each cross*mer cut is placed at the midpoint between the arm's inner gap
+    # edge and the coupler's center-facing boundary (or the arm-tip refpoint when
+    # no coupler is present). This guarantees the cut crosses the gap metal.
+    # Cuts are allowed to overlap each other — not a problem.
+
+    # --- crossleft (West arm) ---
+    # Vertical cut. Inner gap corners: epr_cross_09 (top), epr_cross_06 (bottom).
```
qpavsmi:
```suggestion
    # Cuts for cross partition regions
```
Shorten the comments

klayout_package/python/kqcircuits/simulations/epr/swissmon.py
```diff
@@ -61,70 +118,156 @@ def _get_coupler_wg_offset(i, s):
...
+    cross_xsection_center_top = pya.DPoint(
+        cross_corner_top_l.x + cross_corner_top_w,
+        (cross_corner_top_l.y + coupler_corner_top.y) / 2,
+    )
+    half_cut_length_top = 30.0 + cross_corner_top_w
+
+    result[f"{prefix}crosstopmer"] = {
+        "p1": cross_xsection_center_top + pya.DPoint(-half_cut_length_top, 0),
+        "p2": cross_xsection_center_top + pya.DPoint( half_cut_length_top, 0),
+    }
```
qpavsmi: Could you generalize this part of code into a separate utility function, so that in `correction_cuts` function you could just have four calls:

```
result[f"{prefix}crossleftmer"] = _swissmon_cross_cut(...)
result[f"{prefix}crosstopmer"] = _swissmon_cross_cut(...)
result[f"{prefix}crossrightmer"] = _swissmon_cross_cut(...)
result[f"{prefix}crossbottommer"] = _swissmon_cross_cut(...)
...
```
The arguments for the function could be two `epr_cross` refpoint indices, the appropriate `coupler` refpoint name (or in case of bottom arm cut just another `epr_cross` refpoint). I think that should be enough to return the `"p1", "p2"` dict.

klayout_package/python/kqcircuits/simulations/epr/swissmon.py
qpavsmi: Also move `_offset_point_away` function inside the `partition_regions` function for consistency since `_get_coupler_wg_offset` is already there

qpavsmi (changes requested):
Requesting some changes to make the code easier to read. Please read the suggestions carefully and please manually test that the changes you make don't cause changes to the functionality of swissmon EPR geometry. Looks like after these changes I am ready to merge your submission, but let's see.

AbdullahKazi500:
> Requesting some changes to make the code easier to read. Please read the suggestions carefully and please manually test that the changes you make don't cause changes to the functionality of swissmon EPR geometry. Looks like after these changes I am ready to merge your submission, but let's see.

Ok thanks I will make this changes asap meanwhile you can review my other PR

AbdullahKazi500:
Hi @qpavsmi made the fix there are no breaking changes now ready for review

klayout_package/python/kqcircuits/simulations/epr/swissmon.py
```diff
@@ -61,65 +95,154 @@ def _get_coupler_wg_offset(i, s):
...
+        }
+
 
-    cross_xsection_center = pya.DPoint((cross_corner.x + coupler_corner.x) / 2, cross_corner.y - cross_corner_h)
-    half_cut_length = 30.0 + cross_corner_h
+def correction_cuts(simulation: EPRTarget, prefix: str = "") -> dict[str, dict]:
+    # --- crossleft (West arm) ---
+    # Vertical cut. Inner gap corners: epr_cross_06 (bottom), epr_cross_09 (top).
+    coupler_left = (
+        f"epr_cplr0_max"
```
qpavsmi:
```suggestion
        "epr_cplr0_max"
```
If its a static string then no need for `f` formatter

klayout_package/python/kqcircuits/simulations/epr/swissmon.py
```diff
@@ -61,65 +95,154 @@ def _get_coupler_wg_offset(i, s):
...
-            "p2": cross_xsection_center + pya.DPoint(0, half_cut_length),
-        }
+        f"{prefix}crossleftmer": _swissmon_cross_cut(simulation, 9, 6, coupler_left, axis="x"),
     }
 
-    half_gap = simulation.b / 2
+    # --- crosstop (North arm) ---
+    # Horizontal cut. Inner gap corners: epr_cross_09 (left), epr_cross_00 (right).
+    coupler_top = (
+        f"epr_cplr1_min"
```
qpavsmi:
```suggestion
        "epr_cplr1_min"
```
If its a static string then no need for `f` formatter

klayout_package/python/kqcircuits/simulations/epr/swissmon.py
```diff
@@ -61,65 +95,154 @@ def _get_coupler_wg_offset(i, s):
...
+        if float(simulation.cpl_length[1]) > 0
+        else "epr_cross_10"
+    )
+    result[f"{prefix}crosstopmer"] = _swissmon_cross_cut(simulation, 9, 0, coupler_top, axis="y")
+
+    # --- crossright (East arm) ---
+    # Vertical cut. Inner gap corners: epr_cross_00 (top), epr_cross_03 (bottom).
+    coupler_right = (
+        f"epr_cplr2_min"
```
qpavsmi:
```suggestion
        "epr_cplr2_min"
```
If its a static string then no need for `f` formatter

qpavsmi (changes requested):
This is getting close to merging, I'll just ask you a few small things to do. First is the changes below. Second is that you run following command in terminal, then push the changes:

```
python -m black -l 120 -t py310 klayout_package/python/kqcircuits/simulations/epr/swissmon.py
```

These changes should ensure that this code passes the automatic CI checks

AbdullahKazi500:
> This is getting close to merging, I'll just ask you a few small things to do. First is the changes below. Second is that you run following command in terminal, then push the changes:
> 
> ```
> python -m black -l 120 -t py310 klayout_package/python/kqcircuits/simulations/epr/swissmon.py
> ```
> 
> These changes should ensure that this code passes the automatic CI checks

Ok sure

AbdullahKazi500:
done @qpavsmi

qpavsmi (approved):
Thanks for your contribution, we will merge it soon into the KQC codebase. Hope this was a good learning experience for you on how to work with geometries within KQC code.

AbdullahKazi500:
> Thanks for your contribution, we will merge it soon into the KQC codebase. Hope this was a good learning experience for you on how to work with geometries within KQC code.

yes I want to tackle the next issue now need some guidance

qpavsmi:
I'm closing this PR but give me a few seconds and I will push your changes in a separate way

PR #143, opened 2026-06-08
Add Elmer export macro for layouts drawn in the KLayout GUI

### LINKED ISSUE #122: Export layout present in KLayout GUI as an Elmer simulation

zkasuran:
Closes #122.

Implements the design from issue 122 (Elmer export for layouts drawn in the KLayout GUI).

This adds a macro that takes the geometry currently open in KLayout and exports it as an Elmer wave equation simulation, the same kind of output the existing Python simulation scripts produce.

What it does, following the design in the issue:
- Walks the cell hierarchy of the active design, compounding the instance transforms down the tree.
- Finds Launcher elements and places an EdgePort at each launcher pad edge, at local (s + l, 0).
- Finds qubits, sets their junction_type to Sim and places an InternalPort across the junction from the squid_port_squid_a and squid_port_squid_b reference points.
- Works out the simulation box from the design bounding box and the launcher positions, then projects each EdgePort onto its box edge so the edge ports sit on the rim. It raises a clear error when the launcher orientations leave no room for a valid box.
- Builds the simulation with Simulation.from_cell and exports it with export_elmer and an ElmerVectorHelmholtzSolution.
- Runs on a copy in a new tab so the design you drew is left untouched.

The macro is at scripts/macros/export/export_elmer.lym and mirrors the existing export_ansys.lym. The scraping, port placement and box logic live in a new module simulations/export/elmer/gui_layout_export.py so they can be tested without the GUI. Elements are identified from live PCell data and fall back to the cell name, so the export also works on a flattened layout loaded from a file.

Simulation.from_cell and Simulation.__init__ gain an optional ports argument. The default build does not add ports, so this is the way to attach the ports worked out from the existing geometry. The default is None and existing simulation subclasses are unaffected.

Verification. I do not have Elmer installed, so following the issue I checked the export side only. Running the export headless on the gui_design.oas you attached produces 2 edge ports and 2 internal ports with the gds, json and simulation.sh files. The edge ports land on the box edges. I also ran it on a small design built from live PCell launchers and a Swissmon qubit to exercise the PCell path and the junction_type change. The full test suite passes (473 passed, 11 skipped), black is clean and pylint is 10/10. I added four tests under tests/simulations/test_gui_layout_export.py.

I used the gui_design.oas case you attached, so you can run the macro in the GUI and check the Elmer side. One thing I would like your read on. When launchers on the same side are slightly misaligned I project the edge ports onto a single edge at the outermost launcher. That matches your note that the port can sit a bit off the opening. Let me know if you would rather snap to a different reference.

This change was implemented with substantial help from Claude (Anthropic), which wrote the macro, the module and the tests from the design you laid out in the issue. It was verified locally before submitting: the full test suite (473 passed, 11 skipped), black, pylint 10/10 and the two headless export runs described above. The design choices are documented in the PR for review.

### DIFF
--- a/klayout_package/python/kqcircuits/simulations/export/elmer/gui_layout_export.py
+++ b/klayout_package/python/kqcircuits/simulations/export/elmer/gui_layout_export.py
@@ -1,0 +1,252 @@
+# This code is part of KQCircuits
+# Copyright (C) 2025 IQM Finland Oy
+#
+# This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public
+# License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later
+# version.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied
+# warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along with this program. If not, see
+# https://www.gnu.org/licenses/gpl-3.0.html.
+#
+# The software distribution should follow IQM trademark policy for open-source software
+# (meetiqm.com/iqm-open-source-trademark-policy). IQM welcomes contributions to the code.
+# Please see our contribution agreements for individuals (meetiqm.com/iqm-individual-contributor-license-agreement)
+# and organizations (meetiqm.com/iqm-organization-contributor-license-agreement).
+
+"""Export a manually drawn KLayout layout as an Elmer simulation.
+
+These helpers back the ``export_elmer.lym`` macro. They walk the element hierarchy of a top cell, place
+simulation ports at launchers and qubit junctions, work out the simulation box from the launcher positions,
+and hand the result to :func:`export_elmer`. The logic lives here, separate from the macro, so it can be
+exercised without the KLayout GUI.
+
+Elements are identified both from live PCell data (the usual case when a design is open in the GUI) and, as a
+fallback, from the cell name when the layout has been flattened or loaded from a file without PCell context.
+"""
+
+from pathlib import Path
+
+from kqcircuits.pya_resolver import pya
+from kqcircuits.defaults import default_layers
+from kqcircuits.elements.element import get_refpoints
+from kqcircuits.elements.launcher import Launcher
+from kqcircuits.qubits.qubit import Qubit
+from kqcircuits.simulations.port import EdgePort, InternalPort
+from kqcircuits.simulations.simulation import Simulation
+from kqcircuits.simulations.export.elmer.elmer_solution import ElmerVectorHelmholtzSolution
+from kqcircuits.simulations.export.elmer.elmer_export import export_elmer
+from kqcircuits.simulations.export.elmer.mesh_size_helpers import refine_metal_edges
+
+
+def default_workflow():
+    """Return the default Elmer workflow used by the GUI export."""
+    return {
+        "run_gmsh_gui": True,
+        "run_elmergrid": True,
+        "run_elmer": True,
+        "run_paraview": True,
+        "python_executable": "python",
+        "gmsh_n_threads": -1,
+        "elmer_n_processes": -1,
+        "elmer_n_threads": 1,
+    }
+
+
+def default_mesh_size():
+    """Return the default mesh size dictionary, refined at the metal edges."""
+    return {"global_max": 200.0, **refine_metal_edges(10, 1.0)}
+
+
+def _class_name(cell):
+    """Return the KQC element class name the cell was generated from.
+
+    Cell names look like ``Launcher``, ``Swissmon$1`` or ``Double*Pads``, where ``$`` marks a variant and
+    ``*`` stands for a space. This strips those so the result matches the element class display name.
+    """
+    return cell.name.split("$")[0].replace("*", " ").strip()
+
+
+def _has_sim_junction(cell, layout):
+    """Return True if the cell directly contains a Sim junction instance."""
+    return any(_class_name(layout.cell(inst.cell_index)) == "Sim" for inst in cell.each_inst())
+
+
+def _is_launcher(inst, cell):
+    """Return True if the instance is a Launcher, from PCell data or the cell name."""
+    pcell = inst.pcell_declaration()
+    if pcell is not None:
+        return isinstance(pcell, Launcher)
+    return _class_name(cell) == "Launcher"
+
+
+def _is_qubit(inst, cell, layout):
+    """Return True if the instance is a qubit, from PCell data or a Sim junction child."""
+    pcell = inst.pcell_declaration()
+    if pcell is not None:
+        return isinstance(pcell, Qubit)
+    return _has_sim_junction(cell, layout)
+
+
+def _outward_direction(trans):
+    """Return the global direction of the launcher local +x axis.
+
+    The launcher pad and the chip edge sit on the local +x side, so this vector points outward from the
+    design. It is taken from transformed points rather than the rotation code so it stays correct under
+    mirroring.
+    """
+    return (trans * pya.DPoint(1, 0)) - (trans * pya.DPoint(0, 0))
+
+
+def _launcher_signal_location(inst, cell, trans):
+    """Return the EdgePort location at the launcher pad edge, in global coordinates.
+
+    With PCell data the pad edge is at local ``(s + l, 0)``, the point opposite the launcher opening. For a
+    flat cell the same edge is taken from the local bounding box along +x.
+    """
+    params = inst.pcell_parameters_by_name()
+    if "s" in params and "l" in params:
+        reach = params["s"] + params["l"]
+    else:
+        reach = cell.dbbox().right
+    return trans * pya.DPoint(reach, 0)
+
+
+def _squid_ports(cell, trans, refpoint_layer):
+    """Return the global (signal, ground) points of a qubit Sim junction, or None if absent."""
+    refpoints = get_refpoints(refpoint_layer, cell, pya.DTrans(), None).dict()
+    for name in refpoints:
+        if name.endswith("port_squid_a"):
+            ground = name[:-1] + "b"
+            if ground in refpoints:
+                return trans * refpoints[name], trans * refpoints[ground]
+    return None
+
+
+def collect_ports(top_cell):
+    """Walk ``top_cell`` and build the simulation ports from launchers and qubits.
+
+    Returns a tuple ``(ports, edge_directions)``. ``ports`` is the list of :class:`Port`. ``edge_directions``
+    maps each EdgePort number to its outward :class:`pya.DVector`, used to place that port on the box edge.
+    Launchers and qubits are treated as leaves, every other instance is descended into so elements nested in
+    sub designs are still found.
+    """
+    layout = top_cell.layout()
+    refpoint_layer = layout.layer(default_layers["refpoints"])
+    ports = []
+    edge_directions = {}
+    counter = [1]
+
+    def visit(cell, trans):
+        # Snapshot the instances first. Setting junction_type below regenerates a qubit cell, which would
+        # invalidate a live instance iterator.
+        for inst in list(cell.each_inst()):
+            child = layout.cell(inst.cell_index)
+            child_trans = trans * inst.dcplx_trans
+            if _is_launcher(inst, child):
+                ports.append(EdgePort(counter[0], _launcher_signal_location(inst, child, child_trans)))
+                edge_directions[counter[0]] = _outward_direction(child_trans)
+                counter[0] += 1
+            elif _is_qubit(inst, child, layout):
+                if inst.pcell_declaration() is not None:
+                    inst["junction_type"] = "Sim"
+                    child = layout.cell(inst.cell_index)
+                squid = _squid_ports(child, child_trans, refpoint_layer)
+                if squid is not None:
+                    ports.append(InternalPort(counter[0], squid[0], squid[1]))
+                    counter[0] += 1
+            else:
+                visit(child, child_trans)
+
+    visit(top_cell, pya.DCplxTrans())
+    return ports, edge_directions
+
+
+def simulation_box(top_cell, ports, edge_directions, margin=100.0):
+    """Return the simulation box as a :class:`pya.DBox` and place the edge ports on its rim.
+
+    The box starts from the cell bounding box grown by ``margin``. Each side that faces a launcher is moved
+    to the outermost EdgePort on that side, then every EdgePort on that side is projected exactly onto the
+    edge. Manual placement is rarely pixel perfect, so this projection is what guarantees each EdgePort sits
+    on the box rim, which the Elmer export requires. A clear error is raised when the launcher orientations
+    leave no room for a valid box, for example a port facing right that ends up left of the opposite side.
+
+    The ``signal_location`` of the affected EdgePorts is updated in place.
+    """
+    geometry = top_cell.dbbox()
+    left, bottom = geometry.left - margin, geometry.bottom - margin
+    right, top = geometry.right + margin, geometry.top + margin
+
+    sides = {"left": [], "right": [], "bottom": [], "top": []}
+    for port in ports:
+        direction = edge_directions.get(port.number)
+        if direction is None:
+            continue
+        if abs(direction.x) >= abs(direction.y):
+            sides["right" if direction.x > 0 else "left"].append(port)
+        else:
+            sides["top" if direction.y > 0 else "bottom"].append(port)
+
+    if sides["left"]:
+        left = min(port.signal_location.x for port in sides["left"])
+    if sides["right"]:
+        right = max(port.signal_location.x for port in sides["right"])
+    if sides["bottom"]:
+        bottom = min(port.signal_location.y for port in sides["bottom"])
+    if sides["top"]:
+        top = max(port.signal_location.y for port in sides["top"])
+
+    if right - left <= 0 or top - bottom <= 0:
+        raise ValueError(
+            "Could not define a simulation box from the launcher positions. Check that the launchers point "
+            "outward and sit at the edge of the design."
+        )
+
+    for port in sides["left"]:
+        port.signal_location = pya.DPoint(left, port.signal_location.y)
+    for port in sides["right"]:
+        port.signal_location = pya.DPoint(right, port.signal_location.y)
+    for port in sides["bottom"]:
+        port.signal_location = pya.DPoint(port.signal_location.x, bottom)
+    for port in sides["top"]:
+        port.signal_location = pya.DPoint(port.signal_location.x, top)
+
+    return pya.DBox(left, bottom, right, top)
+
+
+def export_open_layout_to_elmer(
+    top_cell, path, *, name="gui_simulation", target_frequency=5.0, mesh_size=None, workflow=None, margin=100.0
+):
+    """Export the geometry in ``top_cell`` as an Elmer wave equation simulation.
+
+    Args:
+        top_cell: the top cell whose geometry is exported. The cell is modified, so the macro passes a copy.
+        path: directory the simulation files are written to.
+        name: simulation name, also the file name stem.
+        target_frequency: solver frequency in GHz.
+        mesh_size: Gmsh mesh size dictionary, defaults to :func:`default_mesh_size`.
+        workflow: Elmer workflow dictionary, defaults to :func:`default_workflow`.
+        margin: distance in micrometers added around the geometry on sides without a launcher.
+
+    Returns:
+        Path to the written simulation directory.
+    """
+    ports, edge_directions = collect_ports(top_cell)
+    if not edge_directions:
+        raise ValueError("No launchers found in the layout, cannot place edge ports for the simulation.")
+
+    path = Path(path)
+    path.mkdir(parents=True, exist_ok=True)
+    box = simulation_box(top_cell, ports, edge_directions, margin)
+    simulation = Simulation.from_cell(top_cell, name=name, box=box, ports=ports)
+    solution = ElmerVectorHelmholtzSolution(
+        mesh_size=default_mesh_size() if mesh_size is None else mesh_size,
+        frequency=target_frequency,
+        # Use first order basis since the model is very large
+        quadratic_approximation=False,
+        linear_system_method="zmumps",
+        use_multigrid_solver=False,
+    )
+    return export_elmer([(simulation, solution)], path, workflow=default_workflow() if workflow is None else workflow)
--- a/klayout_package/python/kqcircuits/simulations/export/elmer/gui_layout_export.py
+++ b/klayout_package/python/kqcircuits/simulations/export/elmer/gui_layout_export.py
@@ -189,6 +189,11 @@ def collect_ports(top_cell):
     return ports, edge_directions, edge_footprints
 
 
+def _clamp(value, low, high):
+    """Clamp ``value`` to the closed interval ``[low, high]``."""
+    return min(max(value, low), high)
+
+
 def simulation_box(top_cell, ports, edge_directions, edge_footprints, margin=100.0):
     """Return the simulation box as a :class:`pya.DBox` and place the edge ports on its rim.
 
@@ -200,8 +205,9 @@ def simulation_box(top_cell, ports, edge_directions, edge_footprints, margin=100
     guarantees each EdgePort sits on the box rim, which the Elmer export requires.
 
     A clear error is raised when the launcher positions do not admit a valid box: when a side has no room
-    against the opposite side, or when the rim cannot pass through a launcher because it sits too far out
-    relative to its neighbour on the same side.
+    against the opposite side, when the rim cannot pass through a launcher because it sits too far out
+    relative to its neighbour on the same side or when the horizontal and vertical launchers do not agree
+    on one box, so a launcher falls past a box corner and has no point on the finite rim.
 
     The ``signal_location`` of the affected EdgePorts is updated in place.
     """
@@ -236,36 +242,56 @@ def simulation_box(top_cell, ports, edge_directions, edge_footprints, margin=100
             "outward and sit at the edge of the design."
         )
 
-    # Each launcher must still be crossed by its box edge, otherwise it has no point on the rim to host the
-    # port. This rejects a launcher that sits too far out compared to its neighbours on the same side.
+    # Each launcher must sit on its box edge. Two things can go wrong. The box edge may miss the launcher
+    # along its own axis, when the launcher sits too far out compared to its neighbours on the same side.
+    # Or the launcher may fall past a box corner along the other axis, when the horizontal and vertical
+    # launchers do not agree on one box, for example a south launcher pushed out beyond the right edge. Then
+    # there is no point on the finite rim to host the port, so both cases are rejected.
     edges = {"left": left, "right": right, "bottom": bottom, "top": top}
+    too_far_out = (
+        "A launcher sits too far outside the box for the box edge to reach it. The launcher locations are "
+        "incorrect: launchers on the same side should sit at a similar distance from the design."
+    )
+    inconsistent_box = (
+        "The launchers do not fit a single simulation box: a launcher falls past a box corner, so no box "
+        "holds every launcher. Check that the horizontal and vertical launchers all sit at the design edges."
+    )
+
+    # Left and right launchers: the edge must cross the launcher in x. The launcher must also overlap the
+    # vertical span so its port lands on the finite rim, not past a corner.
     for side in ("left", "right"):
         for port in sides[side]:
             footprint = edge_footprints.get(port.number)
-            if footprint is not None and not footprint.left - 1e-6 <= edges[side] <= footprint.right + 1e-6:
-                raise ValueError(
-                    "A launcher sits too far outside the box for the box edge to reach it. The launcher "
-                    "locations are incorrect: launchers on the same side should sit at a similar distance "
-                    "from the design."
-                )
+            if footprint is None:
+                continue
+            if not footprint.left - 1e-6 <= edges[side] <= footprint.right + 1e-6:
+                raise ValueError(too_far_out)
+            if footprint.bottom > top + 1e-6 or footprint.top < bottom - 1e-6:
+                raise ValueError(inconsistent_box)
+
+    # Bottom and top launchers, mirrored: the edge must cross the launcher in y. The launcher must overlap
+    # the horizontal span.
     for side in ("bottom", "top"):
         for port in sides[side]:
             footprint = edge_footprints.get(port.number)
-            if footprint is not None and not footprint.bottom - 1e-6 <= edges[side] <= footprint.top + 1e-6:
-                raise ValueError(
-                    "A launcher sits too far outside the box for the box edge to reach it. The launcher "
-                    "locations are incorrect: launchers on the same side should sit at a similar distance "
-                    "from the design."
-                )
+            if footprint is None:
+                continue
+            if not footprint.bottom - 1e-6 <= edges[side] <= footprint.top + 1e-6:
+                raise ValueError(too_far_out)
+            if footprint.left > right + 1e-6 or footprint.right < left - 1e-6:
+                raise ValueError(inconsistent_box)
 
+    # Project every EdgePort onto its box edge. The free coordinate is clamped into the rim so a launcher
+    # placed slightly past a corner still lands on the box, the same way the edge projection absorbs a small
+    # offset along the launcher's own axis.
     for port in sides["left"]:
-        port.signal_location = pya.DPoint(left, port.signal_location.y)
+        port.signal_location = pya.DPoint(left, _clamp(port.signal_location.y, bottom, top))
     for port in sides["right"]:
-        port.signal_location = pya.DPoint(right, port.signal_location.y)
+        port.signal_location = pya.DPoint(right, _clamp(port.signal_location.y, bottom, top))
     for port in sides["bottom"]:
-        port.signal_location = pya.DPoint(port.signal_location.x, bottom)
+        port.signal_location = pya.DPoint(_clamp(port.signal_location.x, left, right), bottom)
     for port in sides["top"]:
-        port.signal_location = pya.DPoint(port.signal_location.x, top)
+        port.signal_location = pya.DPoint(_clamp(port.signal_location.x, left, right), top)
 
     return pya.DBox(left, bottom, right, top)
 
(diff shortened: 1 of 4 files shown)


zkasuran:
Happy to sign the CLA but the linked form looks broken. The cla-bot link and the one in CONTRIBUTING both redirect to `iqm.tech` after the domain change. That page renders the agreement text under an "IQM Individual Contributor Form" heading but has no name, email or signature fields to fill in, just a newsletter box. So there is no form to submit.

Could you point me to the right place to sign or tell me how you would like me to proceed? Happy to sign by whatever method works on your side.

qpavsmi:
> Happy to sign the CLA but the linked form looks broken. The cla-bot link and the one in CONTRIBUTING both redirect to `iqm.tech` after the domain change. That page renders the agreement text under an "IQM Individual Contributor Form" heading but has no name, email or signature fields to fill in, just a newsletter box. So there is no form to submit.
> 
> Could you point me to the right place to sign or tell me how you would like me to proceed? Happy to sign by whatever method works on your side.

Thank you and apologies, the form broke on our website. You can submit a free-form email to my address: pavel.smirnov@meetiqm.com

In the email please provide your name and email. If you want, you can also provide the company name you are affiliated with (not required).

zkasuran:
Thanks Pavel, that works. I will send my name and email to that address. I will mention both this PR and #144 in the email so the same signature covers the two changes, and I will follow up here once it is sent.

jyesh310:
hello @qpavsmi I wanted to bring something to your attention regarding participation rules for UnitaryHACK. As per the event guidelines, participants are generally expected not to have more than 4 open PRs at the same time.

At the moment, the above participant currently appears to have 7 open PRs across different projects . I thought it would be good to make you aware of this in case it needs review from the maintainers’ side.

I’m only mentioning this to help keep things fair and manageable for both participants and maintainers, especially considering the review workload during the hackathon.

zkasuran:
Sent the CLA details to pavel.smirnov@meetiqm.com just now, with my name and email as requested. It is meant to cover this PR and #144. Ready for verification whenever you get a chance. Thanks.

zkasuran:
Thanks @jyesh310. I am already in contact with the unitaryHACK team about the open-PR count and will be under the limit well within their window. @qpavsmi, sorry for the noise on your thread, this is a hackathon-admin matter and nothing you need to action. Happy to keep this PR focused on the CLA and the code.

jyesh310:
> Thanks @jyesh310. I am already in contact with the unitaryHACK team about the open-PR count and will be under the limit well within their window. @qpavsmi, sorry for the noise on your thread, this is a hackathon-admin matter and nothing you need to action. Happy to keep this PR focused on the CLA and the code.

no issues just wanted to address this as everyone should get a fair chance at the issue thank you

qpavsmi:
@cla-bot check

zkasuran:
Quick housekeeping: I removed the auto-close link to the issue from the description so this PR sits outside the unitaryHACK 4-PR count while the admin side is sorted out, on the unitaryHACK team's suggestion. The PR stays open and the change is unchanged and ready for review. I'll restore the link before merge. Thanks again for the CLA help and for picking this one up.

jyesh310:
> Quick housekeeping: I removed the auto-close link to the issue from the description so this PR sits outside the unitaryHACK 4-PR count while the admin side is sorted out, on the unitaryHACK team's suggestion. The PR stays open and the change is unchanged and ready for review. I'll restore the link before merge. Thanks again for the CLA help and for picking this one up.

Hey I wanted to know how does a 5th PR  works like in your case  outside of the limit I am also interested in a different project and want to make a 5 PR currently I have 3 PRs and want to make two more PRs  thank you

zkasuran:
Hi @qpavsmi, the CLA is signed and the check is green now, so this is ready whenever you have time. A short map to make review quicker.

The macro `scripts/macros/export/export_elmer.lym` is the Elmer counterpart of the existing `export_ansys.lym`: same XML pymacro wrapper, same Export group, "Current view to Elmer". It takes the open design through `KLayoutView(current=True)`, copies it into a new view with `layout.assign` so the original geometry is left untouched, then hands the copy to `export_open_layout_to_elmer`.

The logic sits in `simulations/export/elmer/gui_layout_export.py`, separate from the macro so it can be tested without the GUI. `collect_ports` walks the cell hierarchy and places an `EdgePort` at each launcher pad edge and an `InternalPort` at each qubit squid, identifying elements from PCell data with a cell-name fallback. `simulation_box` fits the box to the launcher positions and projects the edge ports onto the rim, which the export needs. It then goes through the normal `Simulation.from_cell` -> `ElmerVectorHelmholtzSolution` -> `export_elmer` path.

The only change outside the new files is in `simulation.py`: an optional `ports` argument on `__init__` and `from_cell`, defaulting to `None` so existing subclasses that build their own ports are unaffected. It is used here because the ports come from the drawn geometry rather than from `build`.

`tests/simulations/test_gui_layout_export.py` covers the port collection, the box edge projection including the x=0 case, the no-launcher case and a full headless export that asserts the written `sim.json` ports and `sim.gds`. Local run is green: 473 passed, 11 skipped, black clean, pylint 10/10.

Happy to adjust anything to fit your conventions.

veena-vee989:
> Hey I wanted to know how does a 5th PR works like in your case outside of the limit I am also interested in a different project and want to make a 5 PR currently I have 3 PRs and want to make two more PRs thank you

Hi @jyesh310! In this instance there was a technical error on the maintainer/organizer side, which is why we made an exception for this hacker. We appreciate your desire to stick to the rules but ask that you refrain from commenting on others' PRs in the future. If you have any questions, please reach out to hack@unitary.foundation. Thanks!

klayout_package/python/kqcircuits/simulations/export/elmer/gui_layout_export.py
```diff
@@ -0,0 +1,252 @@
+# This code is part of KQCircuits
+# Copyright (C) 2025 IQM Finland Oy
```
qpavsmi:
```suggestion
# Copyright (C) 2026 IQM Finland Oy
```

klayout_package/python/scripts/macros/export/export_elmer.lym
```diff
@@ -0,0 +1,80 @@
...
+ <autorun>false</autorun>
+ <autorun-early>false</autorun-early>
+ <shortcut/>
+ <show-in-menu>false</show-in-menu>
+ <group-name>Export</group-name>
+ <menu-path/>
+ <interpreter>python</interpreter>
+ <dsl-interpreter-name/>
+ <text># This code is part of KQCircuits
+# Copyright (C) 2025 IQM Finland Oy
```
qpavsmi:
```suggestion
# Copyright (C) 2026 IQM Finland Oy
```

tests/simulations/test_gui_layout_export.py
```diff
@@ -0,0 +1,135 @@
+# This code is part of KQCircuits
+# Copyright (C) 2025 IQM Finland Oy
```
qpavsmi:
```suggestion
# Copyright (C) 2026 IQM Finland Oy
```

klayout_package/python/scripts/macros/export/export_elmer.lym
```diff
@@ -0,0 +1,80 @@
...
+path = export_open_layout_to_elmer(
+    copy,
+    path,
+    name=name,
+    target_frequency=target_frequency,
+    mesh_size=mesh_size,
+    workflow=workflow,
+    margin=margin,
+)
+print(f"Exported Elmer simulation to {path}")</text>
```
qpavsmi:
```suggestion
)
preview.layout_view.add_missing_layers()
print(f"Exported Elmer simulation to {path}")</text>
```
This small thing will open the missing simulation layers in the preview layout, I recommend to include this to improve user experience

klayout_package/python/kqcircuits/simulations/export/elmer/gui_layout_export.py
```diff
@@ -0,0 +1,252 @@
...
+
+    Returns:
+        Path to the written simulation directory.
+    """
+    ports, edge_directions = collect_ports(top_cell)
+    if not edge_directions:
+        raise ValueError("No launchers found in the layout, cannot place edge ports for the simulation.")
+
+    path = Path(path)
+    path.mkdir(parents=True, exist_ok=True)
```
qpavsmi: Can you actually check if the exported simulation folder already exists here? Minimum requirement is that if this folder already exists, raise exception and abort the process. A more complete solution would be to invoke [MessageBox](https://www.klayout.de/doc-qt5/code/class_MessageBox.html) to prompt user that the folder exists, and would the user like to delete existing folder to export simulation again, or to abort and leave everything unchanged. I would maybe prefer that the MessageBox handling code would be on the macro side.

klayout_package/python/kqcircuits/simulations/export/elmer/gui_layout_export.py
```diff
@@ -0,0 +1,252 @@
...
+    Returns a tuple ``(ports, edge_directions)``. ``ports`` is the list of :class:`Port`. ``edge_directions``
+    maps each EdgePort number to its outward :class:`pya.DVector`, used to place that port on the box edge.
+    Launchers and qubits are treated as leaves, every other instance is descended into so elements nested in
+    sub designs are still found.
+    """
+    layout = top_cell.layout()
+    refpoint_layer = layout.layer(default_layers["refpoints"])
+    ports = []
+    edge_directions = {}
+    counter = [1]
```
qpavsmi: Why is `counter` a list? Looks like it can just be an integer?

klayout_package/python/kqcircuits/simulations/export/elmer/gui_layout_export.py
```diff
@@ -0,0 +1,252 @@
...
+    box = simulation_box(top_cell, ports, edge_directions, margin)
+    simulation = Simulation.from_cell(top_cell, name=name, box=box, ports=ports)
+    solution = ElmerVectorHelmholtzSolution(
+        mesh_size=default_mesh_size() if mesh_size is None else mesh_size,
+        frequency=target_frequency,
+        # Use first order basis since the model is very large
+        quadratic_approximation=False,
+        linear_system_method="zmumps",
+        use_multigrid_solver=False,
+    )
```
qpavsmi: Could you separate this `solution` into a function similar to `default_workflow` and `default_mesh_size`, for example `default_wave_equation_solution`? Then have `export_open_layout_to_elmer` accept `solution` as argument, which defaults to `None`, in which case it will use `default_wave_equation_solution`.

klayout_package/python/kqcircuits/simulations/export/elmer/gui_layout_export.py
```diff
@@ -0,0 +1,252 @@
...
+    for port in sides["bottom"]:
+        port.signal_location = pya.DPoint(port.signal_location.x, bottom)
+    for port in sides["top"]:
+        port.signal_location = pya.DPoint(port.signal_location.x, top)
+
+    return pya.DBox(left, bottom, right, top)
+
+
+def export_open_layout_to_elmer(
+    top_cell, path, *, name="gui_simulation", target_frequency=5.0, mesh_size=None, workflow=None, margin=100.0
```
qpavsmi:
```suggestion
    top_cell, path, name="gui_simulation", target_frequency=5.0, mesh_size=None, workflow=None, margin=100.0
```
I'm not sure why `*` was there, but if there is good reason then we can keep it

tests/simulations/test_gui_layout_export.py
```diff
@@ -0,0 +1,135 @@
...
+        direction = edge_directions.get(port.number)
+        if direction is None:
+            continue
+        location = port.signal_location
+        if abs(direction.x) >= abs(direction.y):
+            edge = box.right if direction.x > 0 else box.left
+            assert location.x == pytest.approx(edge, abs=1e-6)
+        else:
+            edge = box.top if direction.y > 0 else box.bottom
+            assert location.y == pytest.approx(edge, abs=1e-6)
```
qpavsmi:
```suggestion
    assert set(port.signal_location.x for port in ports) == {box.left, box.right, ???}
```
We have an explicit test example, where we know one signal must be in box left rim, one signal at box right rim, and one signal in the middle. Let's test for that directly, and not have conditionals and part of signal location generation logic in the unit test

qpavsmi (changes requested):
This pull request solves the task almost to completion, well done! I have tested that the exported simulation can be passed to Elmer and properly simulated.

See below some improvements that I'd like to see, and also stay tuned for separate message where I will demonstrate a use case where the `simulation_box` doesn't work 100%

qpavsmi:
There are use cases where I'd like `simulation_box` to work differently. Use case 1:

image

The lower row of launchers are slightly futher back from center compared to upper row launchers. But because the shift is slight, we can still find a box for them. If we push left and right box edges closer to center, then the upper row launchers will be properly at the edge, and the lower row of launchers will be partially outside of the box, but this is fine because there is a point inside launcher that can be on the box rim.

Use case 2 is similar to use case 1, but lower row of launchers are futher apart:

image

Now if you would push the box edges so that upper launchers are on the box edges, then the lower launchers have no point that would also be on box rim. The macro should raise an exception clarifying that the launcher locations are incorrect.

qpavsmi:
Remember to post some message in [issue page](https://github.com/iqm-finland/KQCircuits/issues/122) so that I can put you as an assignee. Also you can probably add the auto-close link, unless you still have PR limit issues

zkasuran:
Thanks for the detailed review, all addressed in the latest push.

- `simulation_box` now follows the launcher closest to the centre on each side, so the launcher nearest the design sits on the rim and a launcher further out only pokes through it (your use case 1). `collect_ports` now returns each launcher footprint. The box raises a clear error when a launcher is too far out for the rim to cross its footprint (use case 2). Two tests cover both cases.
- The library refuses to overwrite an existing output folder (raises `FileExistsError`). The MessageBox handling is on the macro side as you preferred: it asks whether to delete and re-export or abort. On abort it leaves everything unchanged and does not even open a preview tab.
- Added `preview.layout_view.add_missing_layers()` after the export.
- Split the solution into `default_wave_equation_solution`. `export_open_layout_to_elmer` now takes a `solution` argument defaulting to it, matching `default_workflow` and `default_mesh_size`.
- Dropped the `*`. The port counter is a plain integer now and the copyright year is bumped.
- Reworked the box edge test to assert the explicit geometry directly without the direction conditionals.

On the `simulation_box` behaviour I used the launcher footprint (its bounding box) to decide whether the rim still crosses it. If you would rather key the check on a different part of the launcher, point me at it and I will adjust.

zkasuran:
I've added the auto-close link back, so `Closes #122` is on the PR now. The PR-limit issue is sorted on my side, so this one is good to stay linked. I also left a note on #122 so you can assign me there. Thanks again for the thorough review.

qpavsmi:
Another bug with `simulation_box` that I found

image

If I add a south launcher facing away, then it does something strange. There should be some kind of check for both horizontal and vertical facing launchers that they all can be contained within consistant box.

tests/simulations/test_gui_layout_export.py
```diff
@@ -0,0 +1,177 @@
...
+    top.insert(pya.DCellInstArray(qubit.cell_index(), pya.DCplxTrans(1, 0, False, 1500.0, 1500.0)))
+    return layout, top
+
+
+def _cell_with_bbox(left, bottom, right, top_):
+    """Return a layout and a top cell whose bounding box is the given rectangle."""
+    layout = pya.Layout()
+    top = layout.create_cell("Top")
+    top.shapes(layout.layer(1, 0)).insert(pya.DBox(left, bottom, right, top_))
+    return layout, top
```
qpavsmi:
```suggestion
def _cell_with_bbox(x1, y1, x2, y2):
    """Return a layout and a top cell whose bounding box is the given rectangle."""
    layout = pya.Layout()
    top_cell = layout.create_cell("Top")
    top.shapes(layout.layer(1, 0)).insert(pya.DBox(x1, y1, x2, y2))
    return layout, top_cell
```
I suggest this naming, so to not confuse `top` and `top_`

tests/simulations/test_gui_layout_export.py
```diff
@@ -0,0 +1,177 @@
...
+    """The library refuses to overwrite an existing output folder."""
+    layout, top = _design_with_two_launchers_and_qubit()
+    copy_layout = pya.Layout()
+    copy_layout.assign(layout)
+    copy_top = [cell for cell in copy_layout.top_cells() if cell.name == top.name][0]
+
+    out_dir = tmp_path / "sim"
+    out_dir.mkdir()
+    with pytest.raises(FileExistsError):
+        export_open_layout_to_elmer(copy_top, out_dir, name="sim")
```
qpavsmi:
```suggestion
def test_export_raises_when_folder_exists(tmp_path):
    """The library refuses to overwrite an existing output folder."""
    layout, top = _design_with_two_launchers_and_qubit()

    out_dir = tmp_path / "sim"
    out_dir.mkdir()
    with pytest.raises(FileExistsError):
        export_open_layout_to_elmer(top, out_dir, name="sim")
```
I don't think copying layout is needed here

qpavsmi (changes requested):
Just small improvement suggestions in this thread. See also in above comment a bug I found for `simulation_box`, which we could take care of before merging this feature.

zkasuran:
Thanks, both addressed in the latest push (c5c8ba1).

`simulation_box` now reconciles the launchers across both axes. The left and right edges still come from the east-west launchers and the top and bottom from the north-south launchers, but each launcher is now also checked against the other axis. If a launcher's footprint clears a box corner (your south launcher sitting outside the horizontal span) there is no point on the finite rim to host its port, so the export raises a clear error instead of placing the port outside the box. A launcher only slightly past a corner has its port clamped onto the rim, the same way the edge projection already absorbs a small offset along the launcher's own axis. Two tests cover it: a south launcher inside the span lands on the bottom rim, one outside the span raises.

On the clamp: I kept the slightly-past-a-corner case lenient to match how the inner and outer launcher case is handled. The hard error is reserved for a launcher that genuinely cannot share a box. If you would rather it always raise the moment a port would move past a corner, say the word and I will make it strict.

Test tidies: renamed the `_cell_with_bbox` helper params to `x1,y1,x2,y2` so the top cell and the top coordinate no longer clash. Also dropped the layout copy in the folder-exists test.

Verified locally: `tests/simulations` green (75 passed, 7 skipped, the skips need Elmer and Gmsh), pylint 10/10, black clean.

qpavsmi (approved):
Thank you for the implementation! It was a joy to iteratively review the work as it was taking shape!

